PR #15, opened 2017-11-16
Fix login state

SuperManifolds:
Access token is now removed from local storage when user data cannot be retrieved
Access token is now also removed from the server side cookie
Login dialog is presented when user data cannot be retrieved


trezy (approved):
LGTM 👍

PR #16, opened 2017-11-19
Fix login state, the 2nd time around.

UncleClapton:


pages/paperwork/edit.js
```diff
@@ -187,7 +187,7 @@ class Paperwork extends Component {
 
     this.dirtyFields.clear()
 
-    location = `/paperwork/${rescue.id}`
+    Router.push(`/paperwork/${rescue.id}`)
```
trezy: This is how the redirects were done originally. The issue was that when we did the redirect, it would 404 originally, but in dev it would reload moments later when Next polled for changes. Can you verify that doesn't happen anymore?

UncleClapton: Just checked again and this behavior is now back. Not sure why...

Revising this.

UncleClapton:
Final changes for this PR:
1: Move code that was previously in the user reducer to the getUser action method.
2: Fix NaN error in the authentication redirect that occurs on a failed getUser attempt.
3: Invalidate session if either access_token is invalid or missing.
4: Ensure storage access_token is always removed with the cookie access_token.
5: Add myself to the contributors listing.

trezy (approved):
LGTM 👍

PR #18, opened 2017-11-30
Fix various tagsinput bugs.

UncleClapton:
These 2 changes fix warnings being thrown by react due to invalid attribute names and values.

"data-allowDuplicates" and "data-allowNew", while just fine within the context of props, confuses react into thinking they should be data attributes passed to the DOM. Camelcase data attribute names are not valid, and either need to be all lower or separated by a dash. I've opted to make them both lower case, but if you want that changed to dashes I'll change it.

The other issue addressed here involves the input element's autoComplete attribute. autocomplete expects a string value of either "on" or "off", and react doesn't exactly know how to parse a boolean to those values. I've updated the input element in TagsInput to correctly set autocomplete.


UncleClapton:
I'm tagging on another fix to this PR. I found a props/state desync issue with options that effectively breaks the firstlimpet input after it gets manually set once.

trezy (changes requested):
Please make sure this branch is up-to-date with develop. I see a lot of changes that shouldn't be showing up in the PR.

UncleClapton:
Branch updated 👍

trezy (approved):
LGTM 👍

PR #23, opened 2017-12-07
Fix 404 Return error.

UncleClapton:
When attempting to return from a link that leads to a 404 (pressing back), the document is loaded back into the DOM on the same tick as the page itself is. Because the 404 page lacks the head tags defined in the document, the dialogPolyfill isn't defined until after the update tick that calls this componentDidMount function completes. 

... which obviously breaks shit

### DIFF
--- a/components/Head.js
+++ b/components/Head.js
@@ -85,15 +85,12 @@ export default class extends React.Component {
         <meta name="msapplication-wide310x150logo" content="/static/favicon/mstile-310x150.png" />
         <meta name="msapplication-square310x310logo" content="/static/favicon/mstile-310x310.png" />
 
-        <meta name="viewport" content="initial-scale=1.0, width=device-width" />
-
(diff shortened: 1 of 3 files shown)


trezy:
Steps to repro? I think we could probably solve this by moving a lot of things that are normally in `components/Head.js` to `pages/_document.js`. I can do a PoC if you send me steps to repro.

theunkn0wn1:
Go to fuelrats.com and click on profile.
Then click on rats and select one of your rats.

Ofc that page doesn't exist yet so it's a404
Go back on that page and you have bricked the site until manual reload

UncleClapton:
I moved what absolutely needs to be in the main document for things to load properly. Tested working 👍

trezy (approved):
WOOT! 🎉🎊 LGTM 👍

PR #25, opened 2017-12-14
Update site font

UncleClapton:
I'd like to change the main paragraph font to increase readability for blog posts and other smaller text fields.

Some Before/After:

Before:
Before1

After:
After1

Before:

After:

As discussed in IRC, this PR also removes CSS font imports and updates the font header to grab all fonts in a single request to the google font API.

### DIFF
--- a/scss/app.scss
+++ b/scss/app.scss
@@ -1,6 +1,7 @@
 @import 'variables';
 @import 'mixins';
 @import 'colors';
+@import 'fonts';
 
 @import 'reset';
 @import 'root';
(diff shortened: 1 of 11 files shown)


scss/app.scss
```diff
@@ -1,7 +1,6 @@
 @import 'variables';
 @import 'mixins';
 @import 'colors';
-@import 'fonts';
```
trezy: Let's not completely remove `scss/_fonts.scss`. There is the potential that we use a font one day that we need to host ourselves, and I want it to be very clear where that should be done.

trezy (changes requested):

trezy (approved):

UncleClapton:
👍

PR #28, opened 2017-12-22
Localforage migration

UncleClapton:
use localforage for storage as indexeddb seems to be working JUUUUUST dandy.

### DIFF
--- a/components/Page.js
+++ b/components/Page.js
@@ -30,19 +30,14 @@ initStore()
 
 
 
-const dev = preval`module.exports = process.env.NODE_ENV !== 'production'`
-
-
-
-
-
 export default (Component, title = 'Untitled', reduxOptions = {}) => {
   class Page extends React.Component {
     constructor(props) {
       super(props)
 
       LocalForage.config({
-        storeName: dev ? 'FRWebDev' : 'FRWeb',
+        name: 'TheFuelRats',
+        storeName: 'webStore',
       })
     }
 
--- a/components/Page.js
+++ b/components/Page.js
@@ -1,5 +1,6 @@
 // Module imports
 import { bindActionCreators } from 'redux'
+import LocalForage from 'localforage'
 import React from 'react'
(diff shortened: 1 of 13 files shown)


components/Page.js
```diff
@@ -29,8 +30,22 @@ initStore()
...
+
+
+
 export default (Component, title = 'Untitled', reduxOptions = {}) => {
   class Page extends React.Component {
+    constructor(props) {
+      super(props)
+
+      LocalForage.config({
+        storeName: dev ? 'FRWebDev' : 'FRWeb',
```
trezy: Why do we need to namespace the config per environment? Won't they be different just because the URLs are different?

store/reducers/user.js
```diff
@@ -85,9 +79,9 @@ export default function (state = initialState.user, action) {
             user.preferences = user.data.website.preferences
           }
 
-          // Set anything we need in cookies for external access
-          Cookies.set('userId', user.id, dev ? { domain: '.fuelrats.com' } : {})
-          Cookies.set('allowUniversalTracking', user.preferences.allowUniversalTracking, dev ? { domain: '.fuelrats.com' } : {})
```
trezy: I noticed you removed my universal tracking cookie. Add that back, otherwise it's going to cause conflicts with a change I have.

trezy (changes requested):

store/reducers/user.js
UncleClapton: Since we can store the object again, I went back to the preferences object.(see 2 lines below). Do we want to continue with JUST the allowUniversalTracking?

trezy: Yeah, this is fine, then.

components/Page.js
UncleClapton: Ah yeah you're right. This is uneeded. Will revert to a static name

trezy (dismissed):

trezy (approved):
LGTM 👍

PR #29, opened 2017-12-23
Proper null checks are good for you.

UncleClapton:


store/actions/user.js
```diff
@@ -10,7 +10,7 @@ import Router from 'next/router'
 
 // Component imports
 import actionTypes from '../actionTypes'
-
+import initialState from '../initialState'
```
UncleClapton, line 13: is it proper to do this within redux? I'm not sure XD

trezy: Seems like a solid solution to me.

trezy (approved):
LGTM 👍

PR #30, opened 2018-01-03
i-need-fuel page 2.0

UncleClapton:
This is a redo of the i-need-fuel page. The aim is to improve usability and reduce user confusion.

A very common problem with the current page is that the user's eye is drawn towards the green "Chat with the Fuel Rats" button, and clients click that instead of the "get rescued" button. This misdirect is causing clients to land in the casual chatting channel, which is not where client rescues should be held. This issue has been addressed by increasing visibility of the correct button, and pushing the alternate chat button into the background; Away from what the client should be reading.

This new layout also aims to direct non-native English, or non-English, squeakers better. The only element that draws immediate attention from the eye is the large green button in the center of the page. This helps make it clear what button should be pressed to get rescued. Even if the client does not squeak in Modern Galactic English, a large green button is universally known as the "go" button, and should not be mistaken for anything else.

Another issue with the current page is it's large amount of confusing text. Our clients are potentially on emergency oxygen and losing time. Most are inexperienced and panicked about their current situation. Confronting them with a wall of text can not only be confusing, it could mean loss of precious time, or worse, death. Replacing this with some friendly and inviting language, and a notice for CR clients, can help lower overall confusion in what steps to take, as well as reassure the player they are putting their rebuy money in good hands.

In the past, we have had issues with a few clients who have tried to "outsmart" this page by clicking on the chat button instead of the rescue button. They didn't believe their situation was an emergency, when they truly did need fuel. For this reason, language which would imply clicking the "get rescued" button would mean signaling an emergency or priority situation was strictly avoided. Instead, the choices are "get fuel" or "chat with the rats", which will encourage similar minded clients to click the correct button.

Finally, some fine print has been added to the bottom of the page which links to our ToS (and by extension CoC), and PP. This is to increase overall visibility of our policies in case any ToS breaking activity arises.

New I Need Fuel page

### DIFF
--- a/pages/i-need-fuel.js
+++ b/pages/i-need-fuel.js
@@ -1,3 +1,10 @@
+// Module imports
+import Link from 'next/link'
+
+
+
+
+
 // Component imports
 import Page from '../components/Page'
 
@@ -5,7 +12,7 @@ import Page from '../components/Page'
 
 
 
-// Component imports
+// Component constants
 const title = 'I Need Fuel'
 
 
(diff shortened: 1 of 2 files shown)


Belserium:
This certainly looks a good fix. I say release it to live as is, and let's see what kind of effect it has (and then work on it further, if needed).

trezy (approved):
LGTM 👍

PR #31, opened 2018-01-04
Router cleanup

UncleClapton:
This PR cleans up duplicate routes in the router, and captures all methods for 301 and 307 redirect routes.

### DIFF
--- a/server/config/router.js
+++ b/server/config/router.js
@@ -80,7 +80,7 @@ module.exports = function (nextjs, koa, config) {
     await next()
   })
 
-  // Permanent Static Redirects
+  // Permanent Redirects
   router.all('/blogs', async (ctx, next) => {
     ctx.status = 301
     await ctx.redirect(`/blog`)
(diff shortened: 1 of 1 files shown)


server/config/router.js
```diff
@@ -58,33 +58,56 @@ module.exports = function (nextjs, koa, config) {
...
   })
 
-  router.get('/terms-of-service', async (ctx, next) => {
+  router.all('/terms-of-service', async (ctx, next) => {
     ctx.status = 307
     await ctx.redirect(`https://confluence.fuelrats.com/display/FRKB/Terms+of+Service`)
   })
 
-  router.get('/code-of-conduct', async (ctx, next) => {
+  router.all('/code-of-conduct', async (ctx, next) => {
```
trezy, line 110: Why did we change these to `router.all` instead of `router.get`?

UncleClapton: Being that these routes are meant to be redirected, capturing all methods ensures that there isn't any unexpected behavior (in case someone decides to snoop around :P). For these specific routes it really doesn't matter, but it does improve consistency between 301s on nginx and 301s set by the server router. It's more a thinking ahead preventative measure than anything else.

trezy: Fair enough.

trezy (approved):
LGTM 👍

PR #39, opened 2018-01-18
Sidebar link update (landing page), nav.js

Uvelius:
Renaming and completion of the landing page links on fuelrats.com. The new category "Stories, Art, Toons" will be filled though the wordpress interface when ready, and should include all those posts.

### DIFF
--- a/components/Nav.js
+++ b/components/Nav.js
@@ -4,28 +4,22 @@ import React from 'react'
 
 
 
-
-
 // Constants
 const links = {
-Blog: {
-href: '/blog/all',
-as: '/blog',
-},
-Statistics: {
-href: '/statistics',
-},
-Leaderboard: {
-href: '/leaderboard',
-},
-'Stories, Art, & Toons:' {
-href: '/blog/category/138',
-},
-
+  Blog: {
+    href: '/blog/all',
+    as: '/blog',
+  },
+  Statistics: {
+    href: '/statistics',
+  },
+  Leaderboard: {
+    href: '/leaderboard',
+  },
+  'Stories, Art, & Toons:' {
+    href: '/blog/category/138',
+  },
 }
-
-
-
 
 
 export default () => {
--- a/components/Nav.js
+++ b/components/Nav.js
@@ -16,7 +16,7 @@ const links = {
   Leaderboard: {
     href: '/leaderboard',
   },
-  'Stories, Art, & Toons:' {
+  'Stories, Art, & Toons': {
     href: '/blog/category/138',
   },
 }
--- a/components/Nav.js
+++ b/components/Nav.js
@@ -8,7 +8,7 @@ import React from 'react'
 
 // Constants
 const links = {
-  Blog: {
+  News: {
     href: '/blog/all',
     as: '/blog',
   },
@@ -18,6 +18,11 @@ const links = {
   Leaderboard: {
     href: '/leaderboard',
   },
+  Stories, Art, Toons: {
+    href: '/blog/category/138',
+    as: '/blog/category/138',
+  },
+  
 }
 
 
--- a/components/Nav.js
+++ b/components/Nav.js
@@ -8,21 +8,20 @@ import React from 'react'
 
 // Constants
 const links = {
-  News: {
-    href: '/blog/all',
-    as: '/blog',
-  },
-  Statistics: {
-    href: '/statistics',
-  },
-  Leaderboard: {
-    href: '/leaderboard',
-  },
-  Stories, Art, Toons: {
-    href: '/blog/category/138',
-    as: '/blog/category/138',
-  },
-  
+Blog: {
+href: '/blog/all',
+as: '/blog',
+},
+Statistics: {
+href: '/statistics',
+},
+Leaderboard: {
+href: '/leaderboard',
+},
+'Stories, Art, & Toons:' {
+href: '/blog/category/138',
+},
+
 }
 
 


components/Nav.js
```diff
@@ -18,6 +18,11 @@ const links = {
   Leaderboard: {
     href: '/leaderboard',
   },
+  Stories, Art, Toons: {
+    href: '/blog/category/138',
+    as: '/blog/category/138',
```
UncleClapton: What is this "as" here for? Why is it needed?

components/Nav.js
```diff
@@ -18,6 +18,11 @@ const links = {
   Leaderboard: {
     href: '/leaderboard',
   },
+  Stories, Art, Toons: {
+    href: '/blog/category/138',
```
UncleClapton: Are you planning on populating this category? 

I would rather keep stories and art on separate categories and build in a filter to get those specific 2 categories for a proper implementation of what you're looking to achieve here.

Furthermore, category view is terribly broken beyond a single page of results, and needs to be fixed before we can start linking to individual categories.

components/Nav.js
```diff
@@ -8,7 +8,7 @@ import React from 'react'
 
 // Constants
 const links = {
-  Blog: {
+  News: {
```
UncleClapton: Please change this back to "blog" as it's still leading to the blog page. Unless you plan on refactoring half the website to turn "blog" into news, please don't reword just one part of it.

UncleClapton (changes requested):

components/Nav.js
```diff
@@ -18,6 +18,11 @@ const links = {
   Leaderboard: {
     href: '/leaderboard',
   },
+  Stories, Art, Toons: {
```
UncleClapton: This is an invalid name due to how javascript works. You are basically attempting to create a new link called "Toons", and add in 2 existing links called "stories" and "art". That is how the compiler sees it.

UncleClapton (changes requested):

components/Nav.js
Uvelius: Okay, understood

components/Nav.js
Uvelius: Yes, I intended to use this category as a pool for all posts of those three categories. the idea is to save space on the front page by using one link instead of three.

All posts keep their individual categories as well, though. So for the time being, we would have the collective category plus the individual category (f. e. "Art"). We can then look into a filter solution later.

I might be able look into the category display issue later. That has indeed to be fixed first.

components/Nav.js
Uvelius: On an afterthought ... it can be removed indeed.

Uvelius: This would be how it could look like. 

// Constants
const links = {
  Blog: {
    href: '/blog/all',
    as: '/blog',
  },
  Statistics: {
    href: '/statistics',
  },
  Leaderboard: {
    href: '/leaderboard',
  },
  Stories_Art_Toons: {
    href: '/blog/category/138',
  },
  
}

Uvelius (commented):
Possible fixes.

components/Nav.js
UncleClapton: Object property names can be declared as strings for complex names. ```Stories_Art_Toons``` can be ```'Stories, Art, & Toons'```

UncleClapton:
Due to the approach of this solution, I cannot accept this without a fix to the blog's categorical view. The blog currently fails to return results in a sensible manner, and we need a solution to keep it in check.

Uvelius:
Wait ... the category listing is split into several pages, and abbreviated. It's not perfect but seems to work. Example: https://fuelrats.com/blog/category/1   
Isn't it? 
See screenshot: https://imgur.com/OblS15k 

As for the code, this would be 
// Constants
const links = {
Blog: {
href: '/blog/all',
as: '/blog',
},
Statistics: {
href: '/statistics',
},
Leaderboard: {
href: '/leaderboard',
},
'Stories, Art, & Toons:' {
href: '/blog/category/138',
},

}

UncleClapton:
Yes, but look at the results of each page after flipping back and forth between pages a few times. You may see that while results are in order of date, the results themselves seem to be very out of place

Uvelius:
That's weird. Looks like some cache issue or something ... I thought it would be the "endless articles"-problem, but it seems to mix after flipping around.  Though i couldn't observe different categories mixing, more entry dates like.

UncleClapton:
The bug seems to actually be affecting the main blog page as well, so I'm gonna go ahead and add a issue for it on JIRA. See [WEB-151](https://jira.fuelrats.com/browse/WEB-151)

Uvelius:
Oh. Good to know. So it's not just a category issue.

UncleClapton:
I went ahead and fixed the issue with the blog posts. Just took a little troubleshooting to figure out the root cause. All fixed.

Uvelius:
Wow great stuff! Very obliged.  Needs anything else to be addressed with nav.js for merging?

UncleClapton:
@Uvelius please make the changes to the committed code as detailed in the review.

Uvelius:
Done, see https://github.com/Uvelius/fuelrats.com/blob/develop/components/Nav.js

UncleClapton:
Please Update the PR branch so the changes can be pulled into our repo.

Uvelius:
Oh sorry. Have to get used to the workflow first ... https://github.com/FuelRats/fuelrats.com/pull/43

components/Nav.js
```diff
@@ -8,16 +8,20 @@ import React from 'react'
...
-    href: '/blog/all',
-    as: '/blog',
-  },
-  Statistics: {
-    href: '/statistics',
-  },
-  Leaderboard: {
-    href: '/leaderboard',
-  },
+Blog: {
```
UncleClapton: WHOOPS. Forgot to mention :/

Uhh indentation is extremely important to get correct. a single space wrong and the whole project will refuse to build.
each item within a curly brace should be indented 2 spaces more than it's parent.

UncleClapton (changes requested):

Uvelius:
Argh. Of course. I didn't recognize that the spaces have been omitted during code pasting. Sorry. We will get there ...

components/Nav.js
```diff
@@ -18,12 +16,12 @@ const links = {
   Leaderboard: {
     href: '/leaderboard',
   },
+  'Stories, Art, & Toons:' {
```
UncleClapton: Just noticed this as well. the colon needs to be after the closing quote mark for this to be valid code.

UncleClapton (changes requested):

UncleClapton (approved):

UncleClapton:
As Trezy would say, LGTM 👍

Uvelius:
Thanks for your time, that means much to me @UncleClapton  :) I fear I created quite a chaos, which has shown quite clearly that I am not such a big JavaScript wizard (not even apprentice there). I will try to do better contributions in the future ... now that I finally got it that I can contribute to the site development through Github.
Great stuff! Rat fast, fly safe ^, ^/

PR #45, opened 2018-02-26
Include Rat resources overview

Uvelius:
Include a link to our resources overview (links to Reddit, Conf, forum) to make it easier to find all those channels of communication. I hope I didn't mess it up this time ;)

### DIFF
--- a/components/Nav.js
+++ b/components/Nav.js
@@ -16,6 +16,9 @@ const links = {
   Leaderboard: {
     href: '/leaderboard',
   },
+  'Rat links: Forum, Reddit etc.': {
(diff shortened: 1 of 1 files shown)


UncleClapton:
@Uvelius,
I'm rejecting this since the link is to an internal rat resource that should not be made readily available to the general public. We are not posting links to confluence on our sidebar. If it needs to be on the sidebar, the page will need to be on the website itself.

Uvelius:
Okay ...I just run into the problem that our fuelrats.com pages are not really for the public either, only the blog posts. The pages are supposed to be substituted by Confluence as far as I understood it. 
Best way might be a link to the Rat forum thread instead, would it?

PR #50, opened 2018-03-11
2.1.7 - Stability fixes and some new UI

UncleClapton:


components/LoginDialog.js
```diff
@@ -88,7 +106,7 @@ class LoginDialog extends Component {
           <div className="primary">
             <a className="button link" href="/forgot-password">Forgot password?</a>
             <button
-              disabled={!this.state.email || !this.state.password || this.props.loggingIn}
```
itssimple, line 91: Is the removal if `this.state.` intentional?

server/config/router.js
```diff
@@ -33,22 +35,22 @@ module.exports = function (nextjs, koa, config) {
...
-    }
-  })
+  // router.get(authenticatedRoutes, async (ctx, next) => {
+  //   if (ctx.cookie && ctx.cookie.access_token) {
+  //     await next()
+  //
+  //   } else {
+  //     await ctx.redirect(`/?authenticate=true&destination=${encodeURIComponent(ctx.request.url)}`)
+  //   }
+  // })
```
itssimple, line 53: Where did the code for the authenticated routes go? Is this handled by the new router?

server/server.js
```diff
@@ -6,6 +6,9 @@ let isDev = process.env.NODE_ENV !== 'production'
   Module imports
 \******************************************************************************/
 
+// Import variables from .env file.
+require('dotenv').config()
```
itssimple, line 10: If the .env-file should be missing, do we have a graceful fallback, or something that tells us that it's missing?

itssimple (commented):
Looks good, just had some things I wanted to clarify.

components/LoginDialog.js
UncleClapton: yes because email and password are destructured from state on L#56-60

server/config/router.js
UncleClapton: Authenticated routes are defined and handled by the page itself instead of the router now.

See [this](https://github.com/FuelRats/fuelrats.com/pull/50/files/3715894eb2e42aafc508f3026bc6f66d5496df22#diff-2465a0749a2041e522b6b8d7681adf5aR62)

server/server.js
UncleClapton: dotenv just does nothing silently if .env isn't found. dotenv simply exists for dev environment things, and does not affect production or public dev environments.

itssimple (approved):
LGTM

PR #85, opened 2018-06-27
Fix/all the bugs

UncleClapton:
Fixes typo in header
Fixes all redirect flags since a majority of them were broken
Fixes paperwork edit page breaking if system is null.


itssimple (approved):
LGTM 🥇

PR #86, opened 2018-06-27
2.4.1: the update that brought you more authentication redirects and less typos

UncleClapton:


itssimple (approved):
Looks good enough for my machine.

PR #88, opened 2018-08-22
2.4.2 The "some small bits here and there" update, AKA: Clapton keeps doing random things to avoid the bigger project

UncleClapton:
### Changes
* Move register button to below the "get help" button
* Continue making register page more clear that it isn't for clients... because people don't get it.
* Prevent PS4 browser users from raising a r-sig, provide additional instructions on how.
  * kiwi client does not work on PS4 browser
  * It's better to have them move to a phone or computer
* Rewrite various bits of code to keep in line with new rules enforced by airbnb style guides.
  * A few were disabled because even airbnb went back on the specific rules due to how silly they were.
  * Others are generally better for efficiency and play better with the soon-to-be async nature of react.
* Add a first-time welcome message which links new users to the new rat documents.
* Various dependency updates
* Remove subnav state tracking from global state since the sidebar is now a persistent element
* Add "rat links" nav section for logged in rats.
* Fix eslint to allow react fragment shorthand syntax `<>...</>`
* Fix ADD_NICKNAME and CREATE_RAT reducers so they actually update lists properly.
* Add code for deleting nicks, however due to various problems with our current method of nick management, this has been left disabled until we have transitioned to LDAP / Atheme


trezy (changes requested):
The First Login dialog is really hard to read because it has a transparent background. Can we make that not transparent?

PR #91, opened 2018-08-25
Update authorize.js

SuperManifolds:
The logic for the 'strikethrough' text when a permission on the authorise page is inaccessible is incorrect and does the opposite of what it's supposed to, this corrects it.


UncleClapton:
Please resubmit this under the beta branch for current development.

PR #92, opened 2018-08-27
Remove port from rat kiwi

Adanaran:
as the new version runs on default


UncleClapton (approved):
LGTM!

UncleClapton:
Actually.. just noticed this was submitted to the wrong branch. I'm gonna make the change and run it through now

PR #98, opened 2018-10-01
v2.5.0

UncleClapton:
### Major Features
* Storefront + accompanying management pages
* Page locking by API permission feature. AppLayout will display a 401 error page if the current user lacks a specified permission
* Rewritten blog loading logic makes the blog 350% more bearable to load
* Automagically authorize requesting clients if the user has previously accepted access

### Changes
* make various CSS changes to lower space usage in certain cases
* make LoginDialog more mobile friendly
* reword buttons on i-need-fuel __again__ so things are even easier to interpret
* Properly display next/error page when response code is set to an error value
* re-style inputs to glow red when the field is required and empty, green if required and valid
  * This has also been applied to most other custom input fields like TagsInput
* fix authorize page scope list so that accessible client scopes actually display as accessible scopes
* Fix bug in the blog page where removing a filter without reloading the page fully results in the filter staying on
* Improve performance of statistics page by increasing the minimum rescue count to 50 for the rescues by system display
* preload authorize page data so we can remove a loading screen
* No longer assume administrators always have all permissions on the API
* Add 1Password to acknowledgements page.
* Add a list of dependencies to the acknowledgements page

### Tasks
* Migrate to NextJS 7
* Migrate to ESlint 5
* Update ESLint rules, and code to reflect changes in enforced rules.
* Clean up `<AppLayout />` and move user initialization to it's own helper module
* Use dectorators for page flagging and the connect helper function to make our lives a little easier
* rewrite getActionOptions in ActionCreators to make it easier to read and perhaps a bit faster
* remove uglify step in `next.config.js` as next already does that for us on production builds.

I think thats everything ._.


trezy (approved):
LGTM

PR #109, opened 2018-10-11
Release v2.5.1

UncleClapton:
* Move billing info from shipping info stage to order summary page
* Handle cases where the user changes their cart mid-order
* Introduce `sortPriority` metadata field for store products, letting us determine the view order.
* Clean up currency string formation by defining a helper function to do it for us. (Get Money ( ͡° ͜ʖ ͡°))
* Hide attribute descriptors if no attributes for a product exist
* Make it more clear that our billing form is secured through stripe
* Clean up old cruft we don't use
* Remove packages which are no longer used
* Remove all i18n stuff as we have since decided not to translate the website
* Rename instances of `store` subdirectories to `storefront`. This is to avoid confusion between the redux store and the stripe storefront
* Clean up server modules, and only disable certain eslint rules instead of simply turning off eslint entirely for certain modules.
* Clean up getActionOptions even further, and rename to buildActionOptions 
* Forward dispatch and getState functions to action response handlers.
* define an actionStatus enum type.
  * this is currently only being used in the actionCreators file, however the plan is to convert all string comparisons to enum comparisons so eventually we can turn the strings into symbols.
* rewrite blog retrieval logic AGAIN so we can fully remove isomorphic-fetch from the project


helpers/getMoney.js
```diff
@@ -0,0 +1,8 @@
+// ( ͡° ͜ʖ ͡°) because xlexious told me to
```
SuperManifolds, line 1: ( ͡° ͜ʖ ͡°)

server/config/router.js
```diff
@@ -76,47 +39,52 @@ module.exports = function (nextjs, koa) {
...
       day,
       month,
       slug,
       year,
     } = ctx.params
 
-    let dayIsValid = parseInt(day) && (day.length === 2)
-    let monthIsValid = parseInt(month) && (month.length === 2)
-    let yearIsValid = parseInt(year) && (year.length === 4)
+    const dayIsValid = parseInt(day, 10) && (day.length === 2)
```
SuperManifolds, line 49: Why are you using safeParseInt in some places but not here?

SuperManifolds (commented):
Only issue I've found is with the parseInt stuff

server/config/router.js
UncleClapton: cuz thats a server end thing to help resolve old permalinks. I'm not too worried about it as I don't generally expect user-generated input to pass through it. that and this function scares me

SuperManifolds: sounds good to me

helpers/getMoney.js
UncleClapton: ( ͡° ͜ʖ ͡°)

SuperManifolds (approved):
LGTM

PR #130, opened 2018-11-26
feat: Push main site out of the way when opening nav

trezy:
I've moved the #nav-control outside of the header container so that it can affect more than just the
header. I've also updated the transition styles for .user-menu and <main /> so that they smoothly
move out of the way when the nav is opened on mobile.


UncleClapton:
I'm going to close this as I've decided this isn't what we want for the nav on mobile.

PR #135, opened 2019-03-08
Resolved TOS Order for better user experience

EladKarni:
Made it so that TOS only pops up after you check the "I agree". Unchecking it after it has already been checked will uncheck the bools for privacy and terms requiring the user to go through it again.

Example use: https://imgur.com/MMJnKaO

### DIFF
--- a/client/pages/register.js
+++ b/client/pages/register.js
@@ -51,11 +51,13 @@ class Register extends Component {
 
     this._bindMethods([
       'handleChange',
+      'handleTOSChange',
       'handleRadioOptionsChange',
       '_handleSubmit',
     ])
 
     this.state = {
+      checkedTOS: false,
       acceptTerms: true,
       acceptPrivacy: true,
       email: '',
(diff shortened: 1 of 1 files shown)


client/pages/register.js
```diff
@@ -51,11 +51,13 @@ class Register extends Component {
 
     this._bindMethods([
       'handleChange',
+      'handleTOSChange',
```
UncleClapton, line 54: ah lol I didn't realize this component was so far behind...

Not this needs to be changed now... but just so you know class property methods are bound to the instance by default. (the syntax `methodName = () => {...}`) bindMethods pretty obsolete at this point.

UncleClapton (approved):
Looks good to me!

client/pages/register.js
EladKarni: So I know arrow functions have the context set automatically, but you had it one way, and i didn't want to change that. I can make that change fairly quickly if needed

UncleClapton: Nah don't worry about it. We'll tackle it when I come around to cleaning this page up. We have a whole new method for input validation that isn't even implemented on this page yet, and we can do it then

EladKarni: Sounds good!

PR #136, opened 2019-03-09
Redesigned the paperwork view page, included some missing case info.

314numberpi:
Added client name, irc nick, client language, marked for deletion info, board index, timestamps for quotes. Redesigned the page as a whole.


client/pages/paperwork/view.js
```diff
@@ -149,59 +148,90 @@ class Paperwork extends Component {
...
-
-                <tr>
-                  <th>Notes</th>
-                  <td>{rescue.attributes.notes}</td>
-                </tr>
-              </tbody>
-            </table>
+
+            <header className="paperwork-header">
+              {(rescue.attributes.status !== 'open') && (
```
UncleClapton: Shouldn't this be `status !== 'closed'` ?

client/pages/paperwork/view.js
```diff
@@ -149,59 +148,90 @@ class Paperwork extends Component {
...
+              )}
+
+              {(rescue.attributes.status === 'open') && (
+                <div className="tag status-group">
+                  <span className="status">open</span>
+                  <span className="outcome">active</span>
+                </div>
+              )}
+
+              {(rescue.attributes.platform === 'pc') && (
```
UncleClapton: We can also condense this down to 1 block and remove the condition entirely by simply computing the className prop like so:
```jsx
<div className={`tag platform ${rescue.attributes.platform}`}>...</div>
```

client/pages/paperwork/view.js
```diff
@@ -149,59 +148,90 @@ class Paperwork extends Component {
...
+                <span>
+                  Rescue of
+                  <span className="CMDR-name"> {rescue.attributes.client}</span> in
+                  <span className="system"> {(rescue.attributes.system) && (rescue.attributes.system)}</span>
+                </span>
+              </div>
+            </header>
+
+            <div className="tags">
+              {(rescue.attributes.status === 'closed') && (
```
UncleClapton: this status view could be condensed into a single block. status could be computed by just checking if status is closed.
This is where computing some values outside of the render return is useful. We do just that to find out if the user can edit this paperwork.

I would make 2 new variables in the render function like this:
```javascript
// This makes 2 new variables called status and outcome, and sets them to the values of the outcome and status in the rescue object. 
let {
  status,
  outcome,
} = rescue.attributes

if (status === 'inactive') {
  status = 'open'
  outcome = 'inactive'
}

if (status === 'open') {
  outcome = 'active'
}
```

Then in the page markup you can remove the conditionals entirely and just refer to the status and outcome values.

```jsx
<div className="tag status-group">
  <span className="status">{status}</span>
  <span className="outcome">{outcome}</span>
</div>
```

client/pages/paperwork/view.js
```diff
@@ -58,11 +58,10 @@ class Paperwork extends Component {
...
   static renderQuote = (quote, index) => (
     <li key={index}>
-      {quote.message}
-      {Boolean(quote.author) && (
-        <span>
-          - <em>${quote.author}</em>
-        </span>
+      <span className="time">{moment(quote.updatedAt).format('DD MMM, YYYY [\n]HH:mm')}</span>
+      <span className="message">{quote.message}</span>
+      {Boolean(quote.lastAuthor) && (
```
UncleClapton: We should probably still render the original author's name, but if lastAuthor is set perhaps say something like.. `last edited by: {quote.lastAuthor}`.

UncleClapton (changes requested):

client/scss/pages/_paperwork.scss
```diff
@@ -7,9 +7,13 @@
     border-bottom: 1px solid black;
   }
 
-  menu[type=toolbar] div.primary {
-    align-items: center;
+  menu[type=toolbar] {
```
UncleClapton, line 14: I agree that this is quite a bit of CSS, but it's a tough spot we're in since updating old styles might have side effects. I think we should gradually integrate into the other styles as we convert the rest of the site to this new look.

client/pages/paperwork/view.js
314numberpi: Wouldn't it also make sense to keep the original creation date to then? As far as I know the original author is not mentioned outside the API currently.

UncleClapton: given that _all_ case information should be displayed on this page, it's important that we denote that the quote has been edited. Perhaps we should use the `createdAt` time for the main time, and then maybe have some sort of way of showing if it's been edited, with the details appearing in a box on hover.

314numberpi: Will require a bit of a redesign, but as a whole, I agree.

client/pages/paperwork/view.js
314numberpi: Is having the following acceptable to account for unfiled closed cases?
```jsx
{(outcome !== null) && (
  <span className={`outcome ${outcome}`}>{outcome}</span>
)}
```

UncleClapton: I would use a default value if it's unset. is setting the span class in outcome needed for styling? if not I'd just remove it cuz it's not really needed and just use

```jsx
<span className="outcome">{outcome || 'unfiled'}</span>
```

314numberpi: Could have it as 'unfiled'. The span cannot stay empty though as it leaves a box.

client/pages/paperwork/view.js
```diff
@@ -56,16 +56,27 @@ class Paperwork extends Component {
...
-      {quote.message}
-      {Boolean(quote.author) && (
-        <span>
-          - <em>${quote.author}</em>
-        </span>
-      )}
-    </li>
-  )
+  static renderQuote = (quote, index) => {
+    const createdAt = moment(quote.createdAt).format('DD MMM, YYYY HH:mm')
```
UncleClapton: Think we could add 1286 years to this to make it display as 3305?

Since our lint rules don't allow "magic numbers", in other words a number with no attached meaning, you're gonna have to define it as a constant. I would do this outside of the class. There are a couple places we do this already, so probably best to follow the same naming as the others:

```javascript
const ELITE_GAME_YEAR_DISPARITY = 1286
```

client/scss/pages/_paperwork.scss
```diff
@@ -42,5 +46,271 @@
...
+
+        .times {
+          margin-right: 1em;
+
+          .created {
+            margin-right: 0.3em;
+          }
+
+          .updated {
+            &::before {
```
UncleClapton: I'm not really a huge fan of using pseudo-elements for text content like this. it makes it harder to read the render code since it's not readily apparent that this will be the rendered result.

client/pages/paperwork/view.js
```diff
@@ -149,59 +177,68 @@ class Paperwork extends Component {
...
+                  <div className="marked-for-deletion">Marked for Deletion</div>
+                  <div className="md-reason">{rescue.attributes.data.markedForDeletion.reason}</div>
+                  <div className="md-reporter">{rescue.attributes.data.markedForDeletion.reporter}</div>
+                </div>
+              )}
+            </div>
+
+            <div className="info">
+              <span className="label">Created</span>
+              <span className="date-created content">{moment(rescue.attributes.createdAt).format('DD MMM, YYYY HH:mm')}</span>
```
UncleClapton: As above, can we also display this in Elite years?

client/pages/paperwork/view.js
```diff
@@ -149,59 +177,68 @@ class Paperwork extends Component {
...
+              </div>
+            </header>
+
+            <div className="tags">
+              <div className="tag status-group">
+                <span className={`status ${status}`}>{status}</span>
+                <span className="outcome">{outcome || 'unfiled'}</span>
+              </div>
+
+              {(rescue.attributes.platform) && (
```
UncleClapton: So my only reservation about this is that if there is no platform, it's not readily apparent. Obviously once the editor is open it's easy to tell, but not here on the viewing page.

In actual practice... this should be fine as the paperwork editor won't let you file paperwork without a platform, and mecha can't assign without a platform so.. it should be OK?

thoughts?

UncleClapton (changes requested):
I still need to download this to see if I have any issues with the styles yet. You don't happen to have updated screenshots on hand do you? If not I'll just load this myself later today.

client/pages/paperwork/view.js
314numberpi: Same for the info panel, I presume?

UncleClapton: already made a comment about that below 😛

client/pages/paperwork/view.js
314numberpi: How about this?
```
{(rescue.attributes.platform) && (
  <div className={`tag platform ${rescue.attributes.platform}`}>{rescue.attributes.platform}</div>
)}
{(!rescue.attributes.platform) && (
  <div className={`tag platform none`}>No Platform</div>
)}
```
Or is there a better way?

UncleClapton: could remove the conditional all together and do this

```jsx
  <div className={`tag platform ${rescue.attributes.platform || 'none'}`}>{rescue.attributes.platform || 'No Platform'}</div>
```

client/scss/pages/_paperwork.scss
```diff
@@ -42,5 +50,284 @@
...
+          margin-right: 1em;
+
+          .created {
+            margin-right: 1em;
+          }
+
+          .updated {
+            .label {
+              opacity: 0.7;
+              content: ' Updated at ';
```
UncleClapton: did you mean to leave these content properties in here ? 😆

client/pages/paperwork/view.js
```diff
@@ -56,16 +57,27 @@ class Paperwork extends Component {
...
-      {quote.message}
-      {Boolean(quote.author) && (
-        <span>
-          - <em>${quote.author}</em>
-        </span>
-      )}
-    </li>
-  )
+  static renderQuote = (quote, index) => {
+    const createdAt = moment(quote.createdAt).add(ELITE_GAME_YEAR_DESPARITY, 'years').format('DD MMM, YYYY HH:mm')
```
UncleClapton, line 61: I think eventually we're gonna have to write up a file which as all these timestamp formatters so we don't have to keep copypastaing them everywhere. Work for another ticket, though

UncleClapton (changes requested):
See last couple comments :P

client/scss/pages/_paperwork.scss
314numberpi: Nope and they don't do anything anyway!

client/pages/paperwork/view.js
314numberpi: I agree

UncleClapton (approved):

PR #137, opened 2019-03-11
Fix/web 154/whitespace removal

EladKarni:
Added Tooltip for when the pattern doesn't match


client/components/ValidatedFormInput.js
```diff
@@ -1,12 +1,24 @@
...
+import classNames from '../helpers/classNames'
 
 
 
 
 class ValidatedFormInput extends React.Component {
+  /***************************************************************************\
+    Constructor
+  \***************************************************************************/
+  constructor (props) {
```
UncleClapton: constructor is unnecessary here. Prefer using class properties like so:

```javascript
  /***************************************************************************\
    Class Properties
  \***************************************************************************/

  state = {
    errorMessage: '',
  }
```

Also note class structure. 1 blank line between section header and first member of the section, 5 blank lines between last line of code and the next section header. Constructor does not get it's own section if it is needed. It goes in `Public Methods`

client/scss/components/_validated-form-input.scss
```diff
@@ -0,0 +1,44 @@
...
+      pointer-events: none;
+      opacity: 0;
+      max-width: 28rem;
+      background-color: $red;
+      color: $white;
+      text-align: center;
+      border-color: $red-darkened;
+      box-shadow: inset 0 10px 6px -10px rgba(1, 1, 1, 0.5);
+      padding: 1rem 2rem;
+      line-height: 1.25;
```
UncleClapton, line 12: Can we add an opacity transition here? something like....

```sass
transition: opacity .1s ease;
```

UncleClapton (changes requested):

UncleClapton (approved):

PR #152, opened 2019-06-01
feat/WEB-206/Added quotes table

314numberpi:
Rudimentary solution to viewing more case details when editing paperwork


client/pages/paperwork/edit.js
```diff
@@ -479,17 +517,13 @@ class Paperwork extends Component {
...
               <div className="secondary" />
             </menu>
+
+            <div className="panel quotes">
+              <header>Quotes</header>
+              <div className="panel-content">{this.renderQuotes()}</div>
+            </div>
           </form>
         )}
-        <div>
```
UncleClapton, line 484: Ah thanks for taking this out XD

if we're gonna remove the button might as well remove it from state altogether. I'd just remove everything in this component around `userIsCool`

UncleClapton (changes requested):

UncleClapton (approved):
LGTM!

PR #154, opened 2019-06-07
[WEB-217] Carousel image loader worker

EladKarni:
Cleaned up all console logs and finished image loading through web worker

### DIFF
--- a/client/components/Carousel.js
+++ b/client/components/Carousel.js
@@ -2,7 +2,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Transition, animated } from 'react-spring/renderprops.cjs'
-
+import ImageLoaderWorker from '../workers/image-loader.worker'
 
 
 
@@ -57,10 +57,14 @@ class Carousel extends React.Component {
 
   componentDidMount () {
     this.timer = setTimeout(this._setSlide, this.props.interval)
+    this.worker = new ImageLoaderWorker()
+    this.worker.postMessage('from Host')
+    this.worker.addEventListener('message', this.onWorkerMessage)
   }
 
   componentWillUnmount () {
     clearTimeout(this.timer)
+    this.worker.terminate()
   }
 
   renderSlide () {
--- a/client/components/Carousel.js
+++ b/client/components/Carousel.js
@@ -1,11 +1,13 @@
 // Module imports
 import React from 'react'
 import PropTypes from 'prop-types'
+import getConfig from 'next/config'
 import { Transition, animated } from 'react-spring/renderprops.cjs'
 import ImageLoaderWorker from '../workers/image-loader.worker'
 
 
-
+const { publicRuntimeConfig } = getConfig()
+const { publicUrl } = publicRuntimeConfig.local
 
 class Carousel extends React.Component {
   /***************************************************************************\
@@ -14,6 +16,7 @@ class Carousel extends React.Component {
 
   state = {
     curSlide: Object.keys(this.props.slides)[0],
+    imgBlobs: {},
   }
 
   timer = null
@@ -25,6 +28,15 @@ class Carousel extends React.Component {
   /***************************************************************************\
     Private Methods
   \***************************************************************************/
+
+  _handleImageWorkerMessage = (event) => {
+    this.setState((state) => ({
+      imgBlobs: {
+        ...state.imgBlobs,
+        [event.data.id]: [event.data.payload],
+      },
+    }))
+  }
 
   _handleSlideButtonClick = (event) => {
     this._setSlide(event.target.name)
@@ -57,9 +69,10 @@ class Carousel extends React.Component {
 
   componentDidMount () {
     this.timer = setTimeout(this._setSlide, this.props.interval)
+    const imageUrls = Object.entries(this.props.slides).map(([id, slide]) => ([id, `${publicUrl}/static/images/${slide.imageName || `slide_${id}.jpg`}`]))
     this.worker = new ImageLoaderWorker()
-    this.worker.postMessage('from Host')
-    this.worker.addEventListener('message', this.onWorkerMessage)
+    this.worker.postMessage(imageUrls)
+    this.worker.addEventListener('message', this._handleImageWorkerMessage)
   }
 
   componentWillUnmount () {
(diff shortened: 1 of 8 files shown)


client/components/Carousel.js
```diff
@@ -1,11 +1,13 @@
 // Module imports
 import React from 'react'
 import PropTypes from 'prop-types'
+import getConfig from 'next/config'
 import { Transition, animated } from 'react-spring/renderprops.cjs'
+import ImageLoaderWorker from '../workers/image-loader.worker'
```
UncleClapton: Can you separate this out from  the module imports under a new import section commented
`
// Worker imports
`

client/components/Carousel.js
```diff
@@ -1,11 +1,13 @@
...
 
 
-
-
+const { publicRuntimeConfig } = getConfig()
```
UncleClapton, line 11: Please comment this block `// Component constants`

client/components/Carousel.js
```diff
@@ -57,16 +69,24 @@ class Carousel extends React.Component {
 
   componentDidMount () {
     this.timer = setTimeout(this._setSlide, this.props.interval)
+    const imageUrls = Object.entries(this.props.slides).map(([id, slide]) => ([id, `${publicUrl}/static/images/${slide.imageName || `slide_${id}.jpg`}`]))
+    this.worker = new ImageLoaderWorker()
+    this.worker.postMessage(imageUrls)
+    this.worker.addEventListener('message', this._handleImageWorkerMessage)
```
UncleClapton, line 76: Could we add the event listener before communicating with the worker? I know in reality we would never get a response before the listener is added, but I'd like to stick to a logical order of operations for readability.

client/components/Carousel.js
```diff
@@ -57,16 +69,24 @@ class Carousel extends React.Component {
...
     clearTimeout(this.timer)
+    this.worker.terminate()
   }
 
   renderSlide () {
     const {
       slides,
     } = this.props
+    if (typeof this.state.imgBlobs[0] === 'undefined') {
+      return null
```
UncleClapton: We don't need this if we're already null checking in the transition.

client/components/Carousel.js
```diff
@@ -78,19 +98,19 @@ class Carousel extends React.Component {
         leave={{ opacity: 0 }}
         config={{ tension: 280, friction: 85 }}>
         {
-          (curSlide) => (style) => {
+          (curSlide) => this.state.imgBlobs[curSlide] && ((style) => {
```
UncleClapton: perhaps we can destructure `imgBlobs` from `this.state` like we do `slides` from `this.props`, just so we're not calling `this.state` for each reference to `imgBlobs`. easier to read.

client/workers/image-loader.worker.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable no-restricted-globals */
+
+self.addEventListener('message', (event) => {
+  const imageURL = event.data
+  let response = {}
```
UncleClapton: really isn't a point to pre-defining response here. You could remove this and define it as a const in the forEach loop.

client/workers/image-loader.worker.js
```diff
@@ -0,0 +1,16 @@
...
+
+  imageURL.forEach(async (imageData) => {
+    response = await fetch(imageData[1])
+    const blob = await response.blob()
```
UncleClapton: You could also simplify this to something like 

```javascript
const blob = await fetch(imageData[1).then(res => res.blob())
```

Additionally,
if you wanted to make this slightly more readable, you could destructure imageData like:

```javascript
imageURL.forEach(async ([id, url]) => {
  const blob = await fetch(url).then(res => res.blob())
  /* rest of the code */
})
```

client/workers/image-loader.worker.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable no-restricted-globals */
+
+self.addEventListener('message', (event) => {
+  const imageURL = event.data
```
UncleClapton: I'd prefer if we either rename this to imgURLs (plural) or just refereed to event.data directly instead of assigning it to a variable first, since it's only used once.

client/workers/image-loader.worker.js
```diff
@@ -0,0 +1,16 @@
...
+
+  imageURL.forEach(async (imageData) => {
+    response = await fetch(imageData[1])
+    const blob = await response.blob()
+
+    self.postMessage({
+      id: imageData[0],
+      payload: URL.createObjectURL(blob),
+    })
+  })
```
UncleClapton, line 18: Finally.. I think everything within the forEach should be wrapped in a try-catch just so it doesn't break if an image fails to load. Perhaps respond with a null payload if it fails.

UncleClapton (changes requested):

client/components/Carousel.js
EladKarni: Do you still want the event listener in `componentDidMount` just higher up, or somewhere else completely?

client/workers/image-loader.worker.js
EladKarni: So I keep getting an error that `res => res.blob()` should be `(res) => res.blob()`

I thought that if there was only one variable, you could omit the `( )` no?

client/workers/image-loader.worker.js
EladKarni: Should I respond with null or the error itself?

client/workers/image-loader.worker.js
UncleClapton: I made a typo. yes you can omit it, but our code style enforces parens at all times.

client/components/Carousel.js
UncleClapton: no just swap the two lines 😆

client/workers/image-loader.worker.js
UncleClapton: I'd respond with 

```javascript
{
  id,
  payload: null,
}
```
 for now.

We could probably respond with an error in the future once we get some sort of error display alert thing going like we did at RollForGuild, but thats another day.

UncleClapton (approved):
LGTM!

PR #168, opened 2019-06-22
[WEB-211] Account profile redesign

314numberpi:


client/helpers/getDisplayRat.js
```diff
@@ -0,0 +1,20 @@
+const getDisplayRat = (user, ratState) => {
```
UncleClapton: Not sure how this got added back again. File should be removed.

client/components/UserDecalPanel.js
```diff
@@ -87,6 +90,37 @@ class UserDetailsPanel extends Component {
     return null
   }
 
+  _renderDecalCode = (decal, index) => (
+    <>
+      <div className="decal-code" key={`code${index}`}>
+        <button
+          key={`toggle${index}`}
+          aria-label="toggle-decal"
```
UncleClapton: aria-label should be human readable explanatory text, as it is read by screen readers.
```suggestion
          aria-label="Show decal code"
```

client/components/UserDecalPanel.js
```diff
@@ -87,6 +90,37 @@ class UserDetailsPanel extends Component {
     return null
   }
 
+  _renderDecalCode = (decal, index) => (
+    <>
+      <div className="decal-code" key={`code${index}`}>
+        <button
+          key={`toggle${index}`}
```
UncleClapton: unnecessary key being set here. only the root element returned in a map needs the key.

client/components/UserDecalPanel.js
```diff
@@ -87,6 +90,37 @@ class UserDetailsPanel extends Component {
     return null
   }
 
+  _renderDecalCode = (decal, index) => (
+    <>
```
UncleClapton: Avoid using fragments in maps when possible. This should probably be a wrapper `<div />` with an identifying className like `decal-row`

client/components/UserDecalPanel.js
```diff
@@ -87,6 +90,37 @@ class UserDetailsPanel extends Component {
...
+      <div className="decal-code" key={`code${index}`}>
```
UncleClapton: You should never use `index` in a key attribute. 
Since we are mapping out a decal resource, we should use it's id

```suggestion
      <div className="decal-code" key={decal.id}>
```

client/components/UserDecalPanel.js
```diff
@@ -87,6 +90,37 @@ class UserDetailsPanel extends Component {
...
+
+  _renderDecalCodes = () => {
+    const { decals } = this.props
+    console.log(decals)
+
+    if (decals === []) {
+      return null
+    }
+
+    return (
```
UncleClapton: no need for the parenthesis here. you can just do
```javascript
return decals.map(this._renderDecalCode)
```

client/components/UserDecalPanel.js
```diff
@@ -87,6 +90,37 @@ class UserDetailsPanel extends Component {
     return null
   }
 
+  _renderDecalCode = (decal, index) => (
```
UncleClapton: render methods should be "public" (without the `_`), and placed directly above the main render method.

client/components/UserDecalPanel.js
```diff
@@ -87,6 +90,37 @@ class UserDetailsPanel extends Component {
...
+    <>
+      <div className="decal-code" key={`code${index}`}>
+        <button
+          key={`toggle${index}`}
+          aria-label="toggle-decal"
+          className="icon decal-toggle"
+          onClick={this._handleToggle}
+          // title={maxNicksReached ? 'You\'ve used all your nicknames' : 'Add new nickname'}
+          type="button">
+          <FontAwesomeIcon icon={this.state.formOpen[index] ? 'eye-slash' : 'eye'} fixedWidth />
```
UncleClapton: again, avoid using index. Better to use the decal's id like we do with key attributes.

client/components/UserDecalPanel.js
```diff
@@ -87,6 +90,37 @@ class UserDetailsPanel extends Component {
...
+        </button>
+        {decal.attributes.code}
+      </div>
+      <div className="decal-claimed-at" key={`claimedAt${index}`}>{formatAsEliteDate(decal.attributes.claimedAt)}</div>
```
UncleClapton: same suggestion on keys as above, however if we make the wrapper element a div then this can just be removed.

client/components/UserDecalPanel.js
```diff
@@ -108,26 +142,47 @@ class UserDetailsPanel extends Component {
...
+      //   loadingText={checkingEligibility ? 'Checking decal eligibility...' : 'Redeeming decal codes...'}
+      //   manual
+      //   minRows={3}
+      //   noDataText={this._renderNoDataText()}
+      //   showPagination={false} />
+      <div className="panel user-decals">
+        <header>Decal</header>
+        <div className="panel-content">
+          {(checkingEligibility || redeeming) ? (<div className="loading">{loadingText}</div>) : (this._renderDecalCodes(decals))}
+          {(decals === []) ? (this._renderNoDataText()) : ('')}
```
UncleClapton: Same as above, never compare to `[]`

UncleClapton (changes requested):

client/components/UserDecalPanel.js
314numberpi: The problem with that is it messes up the formatting, the grid requires all direct child elements to function correctly (sub-grids, when?), otherwise the date column would either have variable widths for each row, or a fixed-width, not based on the content.

client/components/UserDecalPanel.js
314numberpi: I don't know what you mean by this or why, or even how to do this otherwise?

Wait... length check perhaps...

UncleClapton: wtf... the other one didn't get saved :/

client/components/UserDecalPanel.js
```diff
@@ -87,6 +90,37 @@ class UserDetailsPanel extends Component {
...
+      </div>
+      <div className="decal-claimed-at" key={`claimedAt${index}`}>{formatAsEliteDate(decal.attributes.claimedAt)}</div>
+    </>
+  )
+
+  _renderDecalCodes = () => {
+    const { decals } = this.props
+    console.log(decals)
+
+    if (decals === []) {
```
UncleClapton: Never compare to `[]` as it will always return false. using `[]` always creates a new instance of an object, meaning regardless of contents the two will never be equal. Since decals is always an array, it's safe to do a truthy check on `decals.length`.
```suggestion
    if (!decals.length) {
```

UncleClapton (changes requested):

client/components/AddNicknameForm.js
```diff
@@ -102,7 +122,11 @@ class AddNicknameForm extends Component {
     Redux Properties
   \***************************************************************************/
 
-  static mapDispatchToProps = ['addNickname']
+  static mapDispatchToProps = ['addNickname', 'addNicknamePass']
```
UncleClapton: When `mapDispatchToProps` is given an array of strings, it's telling the `@connect` decorator to map the function of the same name found within the files of the `/client/store/actions` directory. I feel like this whole 'lets define a magic array of strings' thing has been confusing for a lot of people who has seen our website. might just remove it entirely soon once we move to the new namespaced functions thing I keep squeaking about.

Long story short, 'addNicknamePass' is not a valid store function, and there's no reason for it to be there.

client/components/UserDecalPanel.js
```diff
@@ -16,22 +16,21 @@ import Component from './Component'
...
 @connect
 class UserDetailsPanel extends Component {
   /***************************************************************************\
     Class Properties
   \***************************************************************************/
 
   state = {
-    checkingEligibility: true,
-    redeeming: false,
+    ...initialState,
```
UncleClapton: No reason for an `initialState` variable to be defined here as we don't need to clear state for any reason.

client/components/UserDecalPanel.js
```diff
@@ -87,25 +109,31 @@ class UserDetailsPanel extends Component {
...
+      <div className="code">
+        <button
+          aria-label="Show decal code"
+          className="icon toggle"
+          name={decal.id}
+          onClick={this._handleToggle}
+          type="button">
+          <FontAwesomeIcon icon={this.state.decalsVisible[decal.id] ? 'eye' : 'eye-slash'} fixedWidth />
+        </button>
+        {this.state.decalsVisible[decal.id] ? decal.attributes.code : `•••••-•••••-•••••-•••••-${decal.attributes.code.substring(24)}`}
```
UncleClapton: Could we multiline long ternary expressions like this?

```suggestion
        {
          this.state.decalsVisible[decal.id] 
            ? decal.attributes.code 
            : `•••••-•••••-•••••-•••••-${decal.attributes.code.substring(24)}`
        }
```

client/components/UserDecalPanel.js
```diff
@@ -54,13 +53,36 @@ class UserDetailsPanel extends Component {
     })
   }
 
-  static _renderClaimedAtRow (row) {
-    const rescue = row.original
+  _handleToggle = (event) => {
```
UncleClapton: I'd name this something more descriptive like
```suggestion
  _handleDecalVisibilityToggle = (event) => {
```

client/components/AddNicknameForm.js
```diff
@@ -36,24 +39,28 @@ class AddNicknameForm extends Component {
...
+    const {
+      nickname,
+      password,
+    } = this.state
 
     event.preventDefault()
 
     this.setState({ submitting: true })
 
     /* eslint-disable no-alert */
```
UncleClapton: we can remove this eslint disable and enable lines now that we are no longer using a prompt to collect the password.

client/components/ProfileHeader.js
```diff
@@ -0,0 +1,120 @@
...
+import {
+  selectUser,
+  selectUserDisplayRat,
+  selectUserAvatar,
+} from '../store/selectors'
+
+
+
+
+// Component Constants
```
UncleClapton: Could you do a round of clean up on all the files changed/introduced by this PR and remove empty section comments like this?

client/components/ProfileHeader.js
```diff
@@ -0,0 +1,120 @@
...
+        </div>
+      </div>
+    )
+  }
+
+
+
+
+
+  /***************************************************************************\
```
UncleClapton, line 88: empty sections should not have headers like this.

client/components/ProfileHeader.js
```diff
@@ -0,0 +1,120 @@
...
+    Redux Properties
+  \***************************************************************************/
+
+  static mapDispatchToProps = ['getUser']
```
UncleClapton: getUser is unused and not needed here. Further, if the user is already authenticated there is no need to ever call this from a component.

Could just entirely remove this line

client/components/ProfileHeader.js
```diff
@@ -0,0 +1,120 @@
...
+  state = {}
+
+
+
+
+
+  /***************************************************************************\
+    Private Methods
+  \***************************************************************************/
+  _renderUserGroups = () => this.props.user.relationships.groups.data.map((item) => <li className={`badge ${item.id}`} key={item.id}>{item.id}</li>)
```
UncleClapton: Please split this up a little. At least where it becomes JSX so it's not all forced onto one line.
Kind of like:
```suggestion
  _renderUserGroups = () => this.props.user.relationships.groups.data.map((item) => (
    <li className={`badge ${item.id}`} key={item.id}>
      {item.id}
    </li>
  ))
```

Though that might not be enough. If the user has no groups, data will most likely be null. You should probably run a null check before mapping.

client/components/UserDecalPanel.js
```diff
@@ -116,18 +144,16 @@ class UserDetailsPanel extends Component {
...
-        loadingText={checkingEligibility ? 'Checking decal eligibility...' : 'Redeeming decal codes...'}
-        manual
-        minRows={5}
-        noDataText={this._renderNoDataText()}
-        showPagination={false} />
+      <div className="panel user-decals">
+        <header>Decal</header>
+        <div className="panel-content">
+          {(checkingEligibility || redeeming) ? (<div className="loading">{loadingText}</div>) : (this.renderDecalCodes(decals))}
+          {(decals.length) ? ('') : (this.renderNoDataText())}
```
UncleClapton: I would personally write this like:
```suggestion
          {!decals.length && this.renderNoDataText()}
```

React won't render anything if a block evaluates to `false`, `null`, or `undefined`.

client/components/UserDecalPanel.js
```diff
@@ -116,18 +144,16 @@ class UserDetailsPanel extends Component {
...
-        loading={checkingEligibility || redeeming}
-        loadingText={checkingEligibility ? 'Checking decal eligibility...' : 'Redeeming decal codes...'}
-        manual
-        minRows={5}
-        noDataText={this._renderNoDataText()}
-        showPagination={false} />
+      <div className="panel user-decals">
+        <header>Decal</header>
+        <div className="panel-content">
+          {(checkingEligibility || redeeming) ? (<div className="loading">{loadingText}</div>) : (this.renderDecalCodes(decals))}
```
UncleClapton: Another long ternary to be multi-lined

client/components/UserDecalPanel.js
```diff
@@ -139,31 +165,6 @@ class UserDetailsPanel extends Component {
     Getters
```
UncleClapton: Another empty getter section

client/components/UserNicknamesPanel.js
```diff
@@ -4,55 +4,87 @@ import React from 'react'
...
-
+// Component constants
+const MAXNICKS = 20 // Maximum IRC Nicknames allowed
 
 
 
 @connect
 class UserNicknamesPanel extends Component {
-  _deleteNickname = async (event) => {
+  _handleDeleteNickname = async (event) => {
```
UncleClapton: could we restructure this to follow the template, and add in the section headers?

Class Properties (like state)
Private Methods (_ methods/functions)
Public Methods
Getters
Redux Mapping Properties
Prop Definitions

client/components/UserNicknamesPanel.js
```diff
@@ -4,55 +4,87 @@ import React from 'react'
...
     await this.props.deleteNickname(event.target.name)
     return 'Deleted!'
   }
 
+  state = {
+    formOpen: false,
+  }
+
+  _handleToggle = () => {
```
UncleClapton: I'd rename this to something more descriptive like
```suggestion
  _handleFormVisibilityToggle = () => {
```

client/components/UserNicknamesPanel.js
```diff
@@ -4,55 +4,87 @@ import React from 'react'
...
+      ...state,
+      formOpen: !state.formOpen,
+    }))
+  }
+
   render () {
     const {
       user,
     } = this.props
+    const maxNicksReached = (user.attributes.nicknames.length === MAXNICKS)
```
UncleClapton: could we make this `>=` just for safety?

```suggestion
    const maxNicksReached = (user.attributes.nicknames.length >= MAXNICKS)
```

client/components/UserNicknamesPanel.js
```diff
@@ -4,55 +4,87 @@ import React from 'react'
...
                   </ConfirmActionButton>
                   */}
                 </li>
               ))}
           </ul>
         </div>
-        <footer>
-          <AddNicknameForm />
-        </footer>
+        {/* (this.state.formOpen) && (
```
UncleClapton: old garbage?

client/helpers/formatTime.js
```diff
@@ -14,10 +14,13 @@ const formatAsEliteDateTime = (timestamp) => moment(timestamp).add(ELITE_GAME_YE
 
 const formatAsEliteDate = (timestamp) => moment(timestamp).add(ELITE_GAME_YEAR_DESPARITY, 'years').format('DD MMM YYYY').toUpperCase()
 
+const formatAsEliteDateLong = (timestamp) => moment(timestamp).add(ELITE_GAME_YEAR_DESPARITY, 'years').format('DD MMMM YYYY')
```
UncleClapton, line 17: yum more date formatters! 😋

client/components/UserNicknamesPanel.js
```diff
@@ -4,55 +4,87 @@ import React from 'react'
...
     return 'Deleted!'
   }
 
+  state = {
+    formOpen: false,
+  }
+
+  _handleToggle = () => {
+    this.setState((state) => ({
+      ...state,
```
UncleClapton: no need to spread state like this normally. `setState` shallow merges the object you pass it. the only time we need to spread existing state is when we're trying to change something within an object stored within state, without losing the rest of that object.

This `setState` can just become
```javascript
this.setstate((state) => { formOpen: !state.formOpen })
```

client/components/ProfileHeader.js
```diff
@@ -0,0 +1,120 @@
...
+    Public Methods
+  \***************************************************************************/
+
+  render () {
+    const {
+      displayRat,
+      userAvatar,
+    } = this.props
+
+    let { attributes } = this.props.user
```
UncleClapton: These 2 lines could actually be
```suggestion
    const attributes = this.props.user.attributes || {}
```

client/components/UserDecalPanel.js
```diff
@@ -87,25 +109,31 @@ class UserDetailsPanel extends Component {
...
-    } = this.props
-
-    if (!eligible) {
-      await checkDecalEligibility()
+    if (!decals.length) {
+      return null
     }
 
-    this.setState({ checkingEligibility: false })
+    return decals.map(this.renderDecalCode)
```
UncleClapton: you could actually write this bit out as 
```suggestion
    return Boolean(decals.length) && decals.map(this.renderDecalCode)
```
 and remove that `if (!decals.length)` block above.

client/helpers/formatTime.js
```diff
@@ -14,10 +14,13 @@ const formatAsEliteDateTime = (timestamp) => moment(timestamp).add(ELITE_GAME_YE
...
+
 
 
 export {
   // eslint-disable-next-line import/prefer-default-export
```
UncleClapton: Since we have more than one export from this file now, we can remove this config override.

UncleClapton (changes requested):
Here comes the big one 🤣 

Overall really good stuff. I'm loving how this is turning out. As we discussed lets hold off on the last two items for another ticket. I'm going to remove them both and put them into separate tickets.

kenneaal:
Don't mind me, I'm just here to test the announcer suppression.

kenneaal:
> Don't mind me, I'm just here to test the announcer suppression.

Test 2.

kenneaal:
Test 3

kenneaal:
Test 4

kenneaal:
And test 5...

kenneaal:
Say hello to test 6!

kenneaal:
Test 7 has joined the game.

kenneaal:
Test 8 uses PleaseWork.

kenneaal:
It's not very effective.

kenneaal:
Test 9.

kenneaal:
Test 10!

kenneaal:
Why do you hate me, Python?

itssimple:
This is a nice comment. 👌

kenneaal:
Oh, more comments!

kenneaal:
This should not appear.

itssimple:
So fast I can't even!

itssimple:
Fast enough for ya?

kenneaal:
One comment, yes.

kenneaal:
Two comment, yes yes.

client/components/UserDecalPanel.js
```diff
@@ -16,28 +16,11 @@ import Component from './Component'
...
-const ELITE_GAME_YEAR_DESPARITY = 1286 // Years between IRL year and Elite universe year
 
 
 
-/***********************************************/
-/* DUE FOR REFACTORING. DO NOT MODIFY FURTHER. */
-/***********************************************/
 @connect
 class UserDetailsPanel extends Component {
-  /***************************************************************************\
```
UncleClapton, line 28: We still need state, it just doesn't need to be in an initialState variable 🤣

UncleClapton (changes requested):

kenneaal:
These additional testing comments brought to you by Clapton.

kenneaal:
Who is entirely to blame!

UncleClapton (approved):
LGTM!

PR #173, opened 2019-06-27
Fix/add nick password

314numberpi:


UncleClapton (approved):
LGTM, but I think we need to form a proper regex when we come back to this eventually.

PR #179, opened 2019-07-09
Fix/privacy link

noctilucent-dev:
Fixed a couple of broken incorrect links I found (linked to Ts&Cs instead of privacy policy).


UncleClapton (approved):
LGTM!

Thanks for fixing this. I had no idea they were broken. 👍

PR #184, opened 2019-07-15
v2.8.0

UncleClapton:
We are releasing 2.8.0 as-is to push the disabled storefront and our new Teamcity build process. the volume of this update comes from the amount of styling changes made in compliance with our new CSS linter.

## Changes
* Version page no longer links to github commit comparison as that information is not easily accessible via teamcity :(
* Rewrite modals so they have transitions.. and work better.
* Disable storefront while our quartermaster is away.

## Fixes
* Fix Privacy Policy links on `/register` and `/i-need-help`. Thanks @noctilucent-dev 

## Other
* Move website CI to our own teamcity instance. woo!
* We added stylelint to our tool chain, so our css is more betterer.


314numberpi (approved):
Looks good, as far as I can tell.

PR #186, opened 2019-08-18
refactor(password-reset): Added new verify page instead of reset password page

EladKarni:
Created a verify page for handling communication with APIv3 instead of old Reset Password page.


UncleClapton:
Well since it's been so feckin long, Stuff has been kinda been screwed up here. Think this could be rebased to current apiv3-migration and resubmitted?

PR #188, opened 2019-11-03
Fix firefox autofill

UncleClapton:
Yeaaaaaaa validitystate on firefox is a bit crap. checking it again 1ms after the event fire fixes it though. monkaHmm


314numberpi (approved):
LGTM

PR #189, opened 2019-11-03
[WEB-251] don't feed garbage data to system input if stored system is null

UncleClapton:
once again.. monkaHmm


314numberpi (approved):
LGTM

PR #191, opened 2019-11-06
[WEB-235] refactoring password reset (resubmit)

EladKarni:
Resubmitting for verification


UncleClapton (changes requested):
SOOOO Idk what happened here. looks like some autoformatting stuff? Any files touched by your editor seem to get formatted by some extension. ESLint wouldn't enforce these kinds of style rules.

Please revert any files you haven't functionally changed.

client/components/Modal.js
```diff
@@ -39,22 +39,13 @@ const renderModal = ({ item, key, props: style }) => {
...
-      <ModalHeader
-        hideClose={hideClose}
-        onClose={onClose}
-        title={title} />
-
-      <InnerModal {...innerModalProps}>
-        {innerModalChildren}
-      </InnerModal>
-    </RootElement>
+  return (
```
UncleClapton: This is another file that got changed which shouldn't have.

client/store/actionTypes.js
```diff
@@ -52,13 +55,12 @@ const actionTypes = [
...
-].reduce((acc, actionType) => ({
-  ...acc,
-  [actionType]: actionType,
-}), {})
-
-
-
-
+].reduce(
+  (acc, actionType) => ({
```
UncleClapton: This is another change that shouldn't have happened.

client/store/actions/verify.js
```diff
@@ -0,0 +1,22 @@
+import { frApiRequest } from './services'
+import actionTypes from '../actionTypes'
+
+const verifyEmailToken = (token) => frApiRequest(
```
UncleClapton: could you make these inline exports instead of a single export block at the bottom?

so this would become:
```javascript
export const verifyEmailToken = (token) => frApiRequest(
```

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
+/* eslint-disable no-console */
```
UncleClapton: remove this disable directive.

also never disable `no-console`. it's a warning that won't break your build for a reason. `console.log` should only ever be used in debugging and they need to be removed before merging to develop.

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+// Module imports
+import React from 'react'
+// import PropTypes from 'prop-types'
```
UncleClapton: remove unused import please.


also please space the top of this file out. 5 spaces between external and internal imports, and then another 5 between imports and the class.

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+// Component imports
+import { actions, connect, actionStatus } from '../store'
+import { Router, Link } from '../routes'
+import PageWrapper from '../components/PageWrapper'
+import httpStatus from '../helpers/httpStatus'
+import PasswordField from '../components/PasswordField'
+import { passwordPattern } from '../data/RegExpr'
+
+
+// Component Constants
```
UncleClapton: remove unneeded comment. there are no constants

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+    Class Properties
+  \***************************************************************************/
+
+  state = {
+    password: '',
+    submitted: false,
+    submitting: false,
+    validating: false,
+  }
+
```
UncleClapton: please space each class section out.

1 line between the header and the first class member in the section.
5 lines between the end of a section's content and the header of the next.

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+  static async getInitialProps ({ query, store, res }) {
+    // eslint-disable-next-line id-length
+    const { type, t: token } = query
+    let destination = null
+    let response = null
+
+    switch (type) {
+      case 'reset':
+        console.log('Reset Redirect', token)
```
UncleClapton: Remove `console.log` calls

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+        response = await actions.verifyResetToken(token)(store.dispatch)
+        if (response.status === actionStatus.SUCCESS) {
```
UncleClapton: you could check for status above the switch statement instead and skip it entirely if the status is not a success.

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+          destination = '/?authenticate=true'
+        }
+        break
+      default:
+        destination = '/'
+        break
+    }
+    if (destination === null) {
+      return response
```
UncleClapton: you could just wrap the redirect like so instead:

```javascript
if (destination) { 
  if (res) {
    ...
  else {
    ...
  }
}
```

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+      res.writeHead(httpStatus.FOUND, {
+        Location: `${destination}`,
+      })
+      res.end()
+      res.finished = true
+    } else {
+      Router.replace(`${destination}`)
+    }
+
+    return { response }
```
UncleClapton: you should only return what you need when working in `getInitialProps`.

In this case you only need the status so you could pass it back like:
```suggestion
    return {
      tokenIsValid: response.status
    }
```
and then use it as a normal prop.
```javascript
const { tokenIsValid } = this.props
```

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+
+
+
+  /***************************************************************************\
+    Redux Properties
+  \***************************************************************************/
+
+  static mapDispatchToProps = ['resetPassword', 'validatePasswordResetToken']
+
+  static mapStateToProps = () => ({})
```
UncleClapton: since we're not mapping any state. this can be removed

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+
+  /***************************************************************************\
```
UncleClapton: you can remove prop definitions and it's header entirely.

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+    Prop Definitions
+  \***************************************************************************/
+
+  static defaultProps = {}
+
+  static propTypes = {}
+}
+
```
UncleClapton: please space 5 lines between the end of the class and the export line.

UncleClapton (changes requested):

client/pages/verify.js
```diff
@@ -0,0 +1,218 @@
...
+class Verify extends React.Component {
+  /***************************************************************************\
+    Class Properties
+  \***************************************************************************/
+
+  state = {
+    password: '',
+    submitted: false,
+    submitting: false,
+    validating: false,
```
UncleClapton: Whoops missed this one. Validating state can be removed here and where it is checked in `render()`. it's never changed and validation is handled before the page is mounted.

UncleClapton (changes requested):

client/pages/verify.js
EladKarni: Got it! Won't disable it again

client/pages/verify.js
EladKarni: I'm not sure how this would work as they all have different tokens.

client/pages/verify.js
EladKarni: I tried to do what I thought you meant. Let me know if I got it wrong.

EladKarni: This is what I did:
```js    
if (destination) {
      if (res) {
        res.writeHead(httpStatus.FOUND, {
          Location: `${destination}`,
        })
        res.end()
        res.finished = true
      } else {
        Router.replace(`${destination}`)
      }
}
```

EladKarni (commented):
Went one-by-one and commented on each thing and commented on things I wasn't sure about

client/pages/verify.js
UncleClapton: Ah nevermind I was reading the code wrong :P

client/pages/verify.js
```diff
@@ -0,0 +1,228 @@
...
+        }
+        break
+      default:
+        destination = '/'
+        break
+    }
+    if (destination) {
+      if (res) {
+        res.writeHead(httpStatus.FOUND, {
+          Location: `${destination}`,
```
UncleClapton: This string interpolation isn't really needed.

```suggestion
          Location: destination,
```

client/pages/verify.js
```diff
@@ -0,0 +1,228 @@
...
+        })
+        res.end()
+        res.finished = true
+      } else {
+        Router.replace(`${destination}`)
```
UncleClapton: same here
```suggestion
        Router.replace(destination)
```

UncleClapton: I have a function which abstracts this whole `if (res) {...} else {...}` block out. It's not up on the repo yet but when this is merged with those changes we'll probably want to make sure this gets swapped to that.

UncleClapton (changes requested):

UncleClapton:
@all-contributors please add @EladKarni for code and design

PR #198, opened 2019-11-20
[WEB-222] Rewrite RadioOptionsInput

UncleClapton:
# Additions
* Add flavor text to radio options on Paperwork page
* Add link to "How to file cases" confluence article to paperwork edit page

# Changes
* Rewrite `<RadioOptionsInput />` so it doesn't have to use deprecated lifecycle methods \o/
* Removed old checkout code to make way for new stuff later on


314numberpi (approved):
Yep, looks good.

PR #199, opened 2019-11-27
Use Immer to optimize redux state updates, and other various redux changes.

UncleClapton:
### Changes
* Change up vscode launch configurations
* Remove remaining usages of our `React.Component` wrapper, replace them with `React.Component` directly.
    * The use of this wrapper has been unnecessary for a long time, but we haven't removed it until now.
* Add `<React.StrictMode />` to `_app.js` so we know about any bad practices that will impact React's upcoming Concurrent mode
* Use the `immer` library to update redux state. This has simplified our logic quite a bit
* Update the structure of some store members so they're more consistent
* Add new tools for better redux debugging
* Remove nodemon since this seems to break HMR? FeelsBadMan
* Update various dependencies, including the newly released `@fuelrats/next-named-routes` 🎊 

### Fixes
* Fix certain href/as fields in routes so all routes actually work correctly (whoops)


314numberpi (approved):
LGTM, as far as I can tell.

PR #200, opened 2019-12-02
[WEB-218] User Menu

314numberpi:


client/components/UserMenu.js
```diff
@@ -15,84 +15,137 @@ import {
   withCurrentUserId,
 } from '../store/selectors'
 import { Link } from '../routes'
-import AdminUserMenuNav from './AdminUserMenuNav'
 import userHasPermission from '../helpers/userHasPermission'
 
 const UserMenu = (props) => {
+  const {
```
UncleClapton, line 68: Multiple destructures from props :P

could we merge these

client/components/UserMenu.js
```diff
@@ -15,84 +15,137 @@ import {
...
     authenticatedPage,
     loggedIn,
     logout,
     user,
     userAvatar,
     userId,
-    showAdmin,
   } = props
 
+  const renderNavItem = (item) => {
```
UncleClapton: Upon second thought, we should probably split this up a bit for render time efficiency.

could we turn `renderNavItem` and `renderNav` into separate components? Either that or we turn UserMenu into a full class component.

client/components/UserMenu.js
```diff
@@ -15,84 +15,137 @@ import {
...
+      action,
+      className,
+      key,
+      route,
+      routeParams,
+      permission,
+      title,
+    } = item
+
+    if (typeof (permission) === 'undefined' || permission) {
```
UncleClapton: why the explicit check for undefined here?

client/components/UserMenu.js
```diff
@@ -15,84 +15,137 @@ import {
...
+  }
+
+  const renderNav = (nav) => {
+    const {
+      header,
+      items,
+    } = nav
+
+    const permissions = items.map((item) => item.permission)
+    const permitted = permissions.includes(true) || permissions.includes(undefined)
```
UncleClapton: This is kinda inefficient since we have to iterate over every entry 3 times.

perhaps something like:

```javascript
    const permitted = items.filter(({ permission = true }) => permission).length > 0
```

would work better

UncleClapton (changes requested):

UncleClapton (approved):
LGTM!

PR #202, opened 2019-12-08
[WEB-137] [WEB-227] Add nickname reminder

UncleClapton:
* Add reminder that nicknames which are active on irc cannot be registered
* Move some SCSS around so it makes more sense
* start refactoring `<ValidatedFormInput />` as per WEB-227


314numberpi (approved):
Looks great!

PR #205, opened 2019-12-12
2.9 cleanup / update prep

UncleClapton:
* split up and optimize user menu
* adjust user menu transitions
* Clean up some troublesome redux selectors
* Prepare `CHANGELOG` for next version
* Remove teespring link from webstore
* bump version 2.9.0


client/scss/components/_user-menu.scss
```diff
@@ -74,20 +74,17 @@ $user-menu-border-width-mobile: 5px;
     }
 
     @media not screen and (max-width: $mobile-max-width) {
-      right: calc(#{$avatar-width} / 2 + #{$avatar-margin});
```
314numberpi, line 77: Why was this change necessary? If the margin on .user-menu changes, menu will no longer adjust.

314numberpi (changes requested):

client/scss/components/_user-menu.scss
UncleClapton: 😕 

WHHOOOPPPSSSS. That wasn't meant to stay :/

I set it to exact values so I could mess with the position of the menu, but never moved it back :/

client/scss/components/_user-menu.scss
```diff
@@ -178,8 +175,11 @@ $user-menu-border-width-mobile: 5px;
   @media not screen and (max-width: $mobile-max-width) {
     #UserMenuControl:checked {
       ~ menu {
-        max-width: 100vw;
-        max-height: 100vh;
+        top: #{($avatar-size / 2) + 1rem};
```
314numberpi, line 178:
```suggestion
        top: #{($avatar-size / 2) + $avatar-top-margin + 0.5rem};
```

client/scss/components/_user-menu.scss
```diff
@@ -178,8 +175,11 @@ $user-menu-border-width-mobile: 5px;
...
+        right: #{($avatar-size / 2) + 2.5rem};
```
314numberpi, line 179:
```suggestion
        right: #{($avatar-size / 2) + $avatar-right-margin + 0.5rem};
```

314numberpi (changes requested):
I think you would also need to change line 19 of _user-menu.scss to
```
top: $avatar-top-margin;
```
so that it doesn't get left behind.

UncleClapton:
ah yea missed that 😆

314numberpi (approved):
LGTM!

PR #207, opened 2019-12-14
[WEB-260][WEB-261][WEB-262][WEB-263] Fix Various problems intorduced in v2.9.0 (whoops)

UncleClapton:
* Fix decal panel not displaying redeemed decals
* Fix error when logging out from website
* Fix accessToken not being passed to the OAuth authorization page
* Fix registration page improperly checking platform input for value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,18 @@
 
 
 
+## [2.9.1]
+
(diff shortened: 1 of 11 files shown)


314numberpi (approved):
LGTM, as far as I can tell.

PR #210, opened 2020-01-06
[WEB-223] Donation page

UncleClapton:
This PR brings us up to feature parity with the original donation page. New features shall be added at a later date.


client/scss/components/_donate-form.scss
```diff
@@ -0,0 +1,29 @@
...
+    text-align: right;
+  }
+
+  label {
+    text-align: center;
+  }
+
+  .radio-input {
+    justify-content: center;
+  }
```
314numberpi, line 18:
```suggestion
    .option .option-label .option-label-text {
      min-width: 4.5rem
    }
  }
```
Something like this should prevent shuffling between the amount tiers when changing currency. There may be a more elegant solution.

PR #215, opened 2020-01-16
Rework some paperwork stuff (and 1 other left over from another job)

UncleClapton:
### Changes
* Modernizes getFieldValues function for paperwork edit page.
* Moves userCanEdit function into a state selector since thats essentially what it is, and we can even memoize it!
* Moves withStripe decorator into it's own file.

Requesting additional review from @xlexi since the edit permission logic should reflect API behavior. Please see `client/store/selectors/rescues.js`


client/store/selectors/rescues.js
```diff
@@ -1,8 +1,62 @@
...
+      (acc, rat) => {
+        if (rat.attributes.userId === userId) {
+          return [...acc, rat]
+        }
+        return acc
+      }
+    )
+
+    if (usersAssignedRats.length) {
+      return true
```
SuperManifolds, line 45: It should be User is Assigned AND rescue is less than an hour old, not OR

SuperManifolds (changes requested):

client/store/selectors/rescues.js
UncleClapton: huh. I thought it was supposed to be or so that dispatchers could file 🤔

UncleClapton: erm this actually makes no sense... if you're only able to edit if you're assigned AND are within 1h of creation then A LOT of paperwork would never be able to be filed by normal rats :P

SuperManifolds (approved):

PR #216, opened 2020-01-28
v2.10 - Donations and modern code

UncleClapton:
This release is really focused around modernizing some old code thats been laying about. This is the first of a two part release cycle. The second release (v2.10.1) shall finish up the major project wide changes we've been needing to make for a while now, and should be coming within the next few days.

We are releasing this now so that we can get pending functional changes to the website out the door so it's not waiting on many code format changes that will be made in the `v2.10.1` release.

Enjoy!

### ✨ Added
* Add native donation page and it's subsequent donation result page - #210
* Add temporary donation API until APIv3 is out and we can use the new payments API we've been working on that depends on it. - #210
* Add foundation for recaptcha to be used in neccessary places - #210

### ⚡ Changed
* Replace remaining function refs with `React.CreateRef()` - #209
* Further optimize various components by removing inline function attributes (now disallowed by eslint config) - #209
* Re-order imports (enforced by eslint config) - #209
* Move rat tag rendering func to separate helper function. This same code was repeated in various places - #209
* Split router middleware into it's own directory for better route management - #210
* Remove old webstore entirely. Just display disabled option for now showing that it will be back - #210
* Remove antiquated code which used to handle updating blog list page - #210
* Migrate to `next-named-routes@2.0.0` - #215
* Other small refactors for code legibility or optimization - #210, #215

### 🐛 Fixed
* Fix locations of metadata files. They were in static, not on root directory - #210
* Fix disabled nav items so they never link to anything. We just want them sitting there looking pretty - #210
* Fix wp-content proxy origin header setting - #211

### ⚙ Tasks
* Update to Babel 7.8 to support optional chaining and nullish coalescing operators - #216
* Add various new babel transform plugins to support new syntaxes - #209
    * export-namespace-from
    * optional-catch-binding
* Remove do-expressions since they're kinda too new - #216


SuperManifolds (dismissed):
LGTM

SuperManifolds (approved):
LGTM

PR #233, opened 2020-04-13
Fix stuff I broke

UncleClapton:
Heyyyy I broke stuff!

(Imagine that)

### 🐛 Fixed
* Resolved client error that occurs when attempting to authorize an external application - #233
* Resolved server error that arises when donation api bans file is not configured - #233
* Fixed an issue where the description and image associated with a donation amount - #233


314numberpi (approved):
LGTM

PR #238, opened 2020-04-19
[WEB-234] APIv3 request data handling

UncleClapton:
So I've decided we need to commit to APIv3 in full now. As such `apiv3-migration` will be merged after this and we will be working off of develop until APIv3 release.

**NOTE**
Until we reach a certain point of compatibility with APIv3, dev.fuelrats.com will be quite broken :)

This PR adds basic compatibility to all POST/PUT/PATCH requests we send to the API. These are still untested and in some situations the API endpoints might be broken, but the data format should work.


314numberpi (approved):
LGTM

PR #244, opened 2020-05-13
More project fixes and updates

UncleClapton:
* Fix next-redux-wrapper being updated by mistake
* Don't pass the entire error object in back through initialProps because that's bad
* Define a path alias for the `src` directory which can be used anywhere in JS.


314numberpi (approved):
LGTM

PR #245, opened 2020-05-19
Refactor/v3 profile view

314numberpi:
Should allow profile pages to render, which might unblock some tickets.

### DIFF
--- a/src/components/UserNicknamesPanel.js
+++ b/src/components/UserNicknamesPanel.js
@@ -60,7 +60,7 @@ class UserNicknamesPanel extends React.Component {
     const {
       user,
     } = this.props
-    const maxNicksReached = (user.attributes.nicknames.length >= MAXNICKS)
+    const maxNicksReached = (user.relationships.nicknames?.data.length >= MAXNICKS)
 
     return (
       <div className="panel user-nicknames">
@@ -74,7 +74,7 @@ class UserNicknamesPanel extends React.Component {
             )
           }
           <div className="controls">
-            <span className="nickname-count">{`${user.attributes.nicknames.length}/${MAXNICKS}`}</span>
+            <span className="nickname-count">{`${user.relationships.nicknames?.data.length}/${MAXNICKS}`}</span>
             <button
               aria-label="add nickname"
               className={['icon', { green: !this.state.formOpen }]}
@@ -90,17 +90,17 @@ class UserNicknamesPanel extends React.Component {
         <div className="panel-content">
           <ul>
             {
-              user?.attributes.nicknames.map((nickname) => {
+              user?.relationships.nicknames?.data.map((nickname) => {
                 return (
                   <li key={nickname}>
                     <span>{nickname}</span>
-                    {/* <ConfirmActionButton
+                    {/* { <ConfirmActionButton
                       name={nickname}
                       onConfirm={this._handleDeleteNickname}
                       className="icon">
                       <FontAwesomeIcon icon="trash" fixedWidth />
                     </ConfirmActionButton>
-                    */}
+                    } */}
                   </li>
                 )
               }) ?? null
--- a/src/components/UserNicknamesPanel.js
+++ b/src/components/UserNicknamesPanel.js
@@ -60,7 +60,7 @@ class UserNicknamesPanel extends React.Component {
     const {
       user,
     } = this.props
-    const maxNicksReached = (user.relationships.nicknames.data?.length >= MAXNICKS)
+    const maxNicksReached = (user?.relationships.nicknames.data?.length >= MAXNICKS)
 
     return (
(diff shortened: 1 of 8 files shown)


src/components/UserNicknamesPanel.js
```diff
@@ -60,7 +60,7 @@ class UserNicknamesPanel extends React.Component {
     const {
       user,
     } = this.props
-    const maxNicksReached = (user.attributes.nicknames.length >= MAXNICKS)
+    const maxNicksReached = (user.relationships.nicknames?.data.length >= MAXNICKS)
```
UncleClapton: `data` is the nullable here, not `nicknames`.

per JSONAPI spec: https://jsonapi.org/format/#document-resource-object-relationships
> The value of the `relationships` key **MUST** be an object (a “relationships object”).

https://jsonapi.org/format/#document-resource-object-linkage
> Resource linkage MUST be represented as one of the following:
> - `null` for empty to-one relationships.
> - an empty array (`[]`) for empty to-many relationships.
> - a single resource identifier object for non-empty to-one relationships.
> - an array of resource identifier objects for non-empty to-many relationships.


Also, user should be considered nullable :)

src/components/UserNicknamesPanel.js
```diff
@@ -74,7 +74,7 @@ class UserNicknamesPanel extends React.Component {
             )
           }
           <div className="controls">
-            <span className="nickname-count">{`${user.attributes.nicknames.length}/${MAXNICKS}`}</span>
+            <span className="nickname-count">{`${user.relationships.nicknames?.data.length}/${MAXNICKS}`}</span>
```
UncleClapton: same as above.

src/components/UserNicknamesPanel.js
```diff
@@ -90,17 +90,17 @@ class UserNicknamesPanel extends React.Component {
         <div className="panel-content">
           <ul>
             {
-              user?.attributes.nicknames.map((nickname) => {
+              user?.relationships.nicknames?.data.map((nickname) => {
```
UncleClapton: same as above.

src/store/actions/groups.js
```diff
@@ -0,0 +1,19 @@
+import actionTypes from '../actionTypes'
+import { frApiRequest } from './services'
+
+export const getGroups = (params) => {
```
UncleClapton, line 4: There shouldn't really be a need to get groups since they're always included with a user, but sure we can add these for completeness.

Also the action type for `getGroups` should be `groups.search` and not `groups.read`. `read` is for a single resource and `search` is for listing endpoints. This mirrors terminology used on the API.

src/store/selectors/groups.js
```diff
@@ -28,7 +28,7 @@ export const selectGroups = (state) => {
 
 
 export const selectGroupById = (state, { groupId }) => {
-  return state.groups[groupId]
+  return state.groups?.[groupId]
```
UncleClapton: the groups object itself should always be defined. if its not then we need to fix initialState.

src/store/selectors/decals.js
```diff
@@ -31,7 +31,7 @@ export const selectDecalsByUserId = createSelector(
   [selectDecals, getUserId],
   (decals, userId) => {
     return Object.values(decals).filter((decal) => {
-      return decal.attributes && decal.attributes.userId === userId
+      return decal?.attributes?.userId === userId
```
UncleClapton, line 34: uhh I'm gonna accept this as-is for now, but these null checks should no longer be required after the decal state object stops tracking if the user is eligible.

UncleClapton (changes requested):

src/store/selectors/decals.js
314numberpi: Yeah, it simply so it'll load. Definitely should be revisted.

src/store/selectors/groups.js
314numberpi: I may have been overzealous here.

UncleClapton: we should only null check what can be null :P

src/components/UserNicknamesPanel.js
```diff
@@ -60,7 +60,7 @@ class UserNicknamesPanel extends React.Component {
     const {
       user,
     } = this.props
-    const maxNicksReached = (user.relationships.nicknames?.data.length >= MAXNICKS)
+    const maxNicksReached = (user.relationships.nicknames.data?.length >= MAXNICKS)
```
UncleClapton: as mentioned before, user should still be considered nullable. while technically it shouldn't it's worth being careful about it because of page change edge cases.

UncleClapton (changes requested):

UncleClapton (approved):
lgtm

PR #246, opened 2020-05-20
This is the one where I PR a bunch of stuff that has nothing to do with the branch name I'm PRing it from, but we want these changes in develop now so others can use it and I committed it all in this branch so deal with it.

UncleClapton:
dealwithit


314numberpi (approved):
O.o

PR #248, opened 2020-05-21
[WEB-231][WEB-255] rewrite add nickname form

314numberpi:
Nicks can't be added at the moment, pending WEB-255


src/components/AddNicknameForm/AddNicknameForm.js
```diff
@@ -75,47 +74,35 @@ class AddNicknameForm extends React.Component {
...
           name="nickname"
           pattern={ircNickPattern}
           patternMessage="Nickname must start with a letter, contain no spaces, and be between 2-30 characters"
           placeholder="Add a nickname..."
+          title={this.props.title}
           type="text"
           value={nickname}
           onChange={this._handleChange}>
-          <InfoBubble header="reminder" id="NickRegisterReminder">
+          <InfoBubble className={bubbleStyles.nicknameInfo} header="reminder" id="NickRegisterReminder">
```
UncleClapton: `nicknameInfo` should be a part of `AddNicknameForm`'s styles if it's only used by this component.

UncleClapton (changes requested):

src/components/AddNicknameForm/AddNicknameForm.js
314numberpi: That is much simpler :P

src/components/AddNicknameForm/AddNicknameForm.js
```diff
@@ -44,18 +40,16 @@ class AddNicknameForm extends React.Component {
...
       nickname,
-      password,
     } = this.state
 
     event.preventDefault()
 
     this.setState({ submitting: true })
 
-    await addNickname(user.id, nickname, password)
+    await addNickname(user, { attributes: { nick: nickname } })
```
UncleClapton: so. We should really have some form of error handling on this in case things break.

src/components/UserNicknamesPanel/UserNicknamesPanel.js
```diff
@@ -44,7 +54,9 @@ class UserNicknamesPanel extends React.Component {
   }
 
   _handleDeleteNickname = async (event) => {
-    await this.props.deleteNickname(event.target.name)
+    await this.props.deleteNickname(this.props.user, this.props.nicknames.find((nick) => {
```
UncleClapton: We should check the response for an error here and display accordingly.

use `isError()` from `flux-standard-action` to check the result from the action now.

UncleClapton (changes requested):

src/components/AddNicknameForm/AddNicknameForm.js
```diff
@@ -41,12 +44,40 @@ class AddNicknameForm extends React.Component {
...
+
+      if (HttpStatus.isClientError(meta.response.status)) {
+        errorMessage = payload.errors && payload.errors.length ? payload.errors[0].title : 'Client communication error'
+      }
+
+      if (HttpStatus.isServerError(meta.response.status)) {
+        errorMessage = 'Server communication error'
+      }
+
+      if (meta.response.status === httpErrors.conflict) {
```
UncleClapton: HttpStatus has all of these defined, no need to define them again in the object above.

https://github.com/FuelRats/fuelrats.com-utils/blob/master/src/http/index.js#L120

UncleClapton (changes requested):

UncleClapton (approved):

UncleClapton:
LGTM!

PR #251, opened 2020-06-11
Fix data handling (holy crap)

UncleClapton:
soooo I completely rewrote stuff and ended up with something that brings us in line with plans for FRAPIccino... mostly.


314numberpi (approved):
Ouch.

UncleClapton:
ESLint errors are fine. They are TODOs which will be handled in my next PR

PR #255, opened 2020-07-15
[WEB-233] APIv3 Leaderboard

UncleClapton:
* Fixes leaderboard and supporting global state code so it works with APIv3

This is a simple patch for compatibility. We still intend on re-designing the leaderboard in the future.


src/components/RatLeaderboardTable.js
```diff
@@ -125,16 +119,16 @@ class RatLeaderboardTable extends React.Component {
...
               rescueCount,
             } = value.attributes
 
             return (
               <div className="badge-list">
                 <RescueAchievementIcon className="size-32 fixed" rescueCount={rescueCount} />
-                <CodeRedIcon className="size-32 fixed" codeRedCount={safeParseInt(codeRed)} />
-                <FirstYearIcon className="size-32 fixed" createdAt={createdAt} />
+                <CodeRedIcon className="size-32 fixed" codeRedCount={codeRedCount} />
+                <FirstYearIcon className="size-32 fixed" createdAt={joinedAt} />
```
314numberpi, line 131: Wouldn't it make sense to change `createdAt` to `joinedAt` as well? Seems more legible.

src/store/initialState.js
```diff
@@ -24,8 +24,8 @@ const initialState = {
   images: {},
 
   leaderboard: {
-    loading: false,
```
314numberpi, line 27: I do not see why `loading` is no longer required in the initial state?

UncleClapton: Because we track loading locally within the table

src/components/RatLeaderboardTable.js
UncleClapton: yeaaaaaaa I just wasn't too worried about it because we're gonna be ditching most of this someday soon ™

314numberpi (approved):
LGTM

PR #260, opened 2020-07-17
[WEB-284] Update decals for APIv3

314numberpi:


src/store/selectors/decals.js
```diff
@@ -28,10 +21,13 @@ export const selectDecalById = (state, { decalId } = {}) => {
...
-      return decal?.attributes?.userId === userId
-    })
+  [selectDecals, selectUserById],
+  (decals, user) => {
+    if (user) {
+      return user.relationships.decals?.data.map(({ id }) => {
+        return decals[id]
+      })
+    }
+    return []
```
UncleClapton: We should prefer to return `undefined` if no user is found since there is a difference between there being no decals for the user and no user at all.

UncleClapton (changes requested):

UncleClapton (approved):
LGTM!

PR #261, opened 2020-07-18
Fix redux middleware

UncleClapton:
There was an error with our new error handling middleware which this patch resolves (no the irony has not been lost on me). I have also moved both of our custom middleware to our utility package, which can be viewed [here](https://github.com/FuelRats/fuelrats.com-utils/blob/master/src/redux-middleware.js)

I also refactored `initUserSession()` a bit so it's simpler


314numberpi (approved):
LGTM

PR #263, opened 2020-07-30
[Untracked][WEB-278] APIv3 password resets

UncleClapton:
This PR finalizes support for password resets through APIv3. It also introduces a redesigned `<LoginModal />`

* Bunch of little behavior tweaks to `useForm()` (more magic!)
* rewritten `<LoginForm />` which splits the component up into 3 different views
* Move to next.js built in dotenv loader for dev mode (everyone will need to rename `.env` to `.env.local` eventually, but right now both should work just fine)
* Rewrite `<Switch />`
* A bunch of other small stuff I really can't be bothered to list
* Upgrade to Next.js 9.5 (see more [here](https://nextjs.org/blog/next-9-5))

(I'll try to actually write up a comprehensive list of changes tomorrow, but for now I'm sleeping)

this also fixes WEB-278.


314numberpi (changes requested):
- ~~[RatLeaderBoardTable.js#L58](https://github.com/FuelRats/fuelrats.com/pull/263/files#file-src-components-ratleaderboardtable-js-L58) - Not sure why this was missed/left~~

- [RegExpr.js#L4](https://github.com/FuelRats/fuelrats.com/blob/153ec4cc0b0d52cc5e24c84f2023c46345046079/src/data/RegExpr.js#L4) - Seems as though this is unused now, or at least is not in-line with the password validation in PasswordFieldset

- [InputFieldset.js#L89](https://github.com/FuelRats/fuelrats.com/blob/153ec4cc0b0d52cc5e24c84f2023c46345046079/src/components/Fieldsets/InputFieldset/InputFieldset.js#L89) - Are we still planning on using fieldsets for these (as per discussion with Trezy)? Same on SwitchFieldset and RadioFieldset

- [IRCNickFieldset.js#L36](https://github.com/FuelRats/fuelrats.com/blob/153ec4cc0b0d52cc5e24c84f2023c46345046079/src/components/Fieldsets/IRCNickFieldset.js#L36) - We occasionally get a rat with 'guild' tags in their CMDR name. Could/should we warn on this rather than block validation?

UncleClapton:
> Seems as though this is unused now, or at least is not in-line with the password validation in PasswordFieldset

It's still used in `<ChangePasswordModal />` I considered refactoring that to be slightly too far out of scope for this PR.

> Are we still planning on using fieldsets for these (as per discussion with Trezy)? Same on SwitchFieldset and RadioFieldset

No I just CBA to change them and deal with the styling of a div to make it behave the same at the moment. I kinda considered it a change for after everything is ready to be pushed to release.

> We occasionally get a rat with 'guild' tags in their CMDR name. Could/should we warn on this rather than block validation?

Guild tags in their CMDR name is fine, but I think they should be avoided in IRC nicks. We could bring this up as a topic in the upcoming sync meeting and see what others think.

314numberpi (approved):

PR #298, opened 2020-12-02
Sort cases with "last created to first created" on initial load.

diraven:


UncleClapton:
Hey there. Could you add the change to our changelog?

CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ For detailed rules of this file, see  [Changelog Rules](#changelog-rules)
 
 
 ### 🐛 Fixed
-*
+* Change cases sorting to "last created first" on initial dispatch board load.
```
UncleClapton: I'm going to be adding this to `CONTRIBUTING.md` so there's no confusion going forward, but changes should link back to the PR that introduced them.

In this case you need to add the PR# like this:
```markdown
* Change cases sorting to "last created first" on initial dispatch board load. - [#298][]
```

Then in the section between the unreleased changes and the previous release there are link definitions. It's here that you'll want to add:

```markdown
[#298]: https://github.com/fuelrats/fuelrats.com/pull/297
```

UncleClapton (changes requested):

CHANGELOG.md
diraven: Just curious - why not regular md links like `[link to Google!](http://google.com)`?

UncleClapton (approved):
LGTM!

CHANGELOG.md
UncleClapton: There are often cases where there are multiple entries for a single PR. In that case it's better to just define the link once for cleanliness. To be consistent with those cases, we always just define links like this.

UncleClapton:
@all-contributors please add @diraven for code

PR #305, opened 2021-01-03
WEB-283 Logging out revokes your access_token through API

Master-Guy:
Depends on https://github.com/FuelRats/api.fuelrats.com/pull/598

### DIFF
--- a/src/store/actions/session.js
+++ b/src/store/actions/session.js
@@ -1,11 +1,13 @@
-import { createFSA } from '@fuelrats/web-util/actions'
+import { createFSA, createAxiosFSA } from '@fuelrats/web-util/actions'
 import { HttpStatus } from '@fuelrats/web-util/http'
 import { isError } from 'flux-standard-action'
 
 import { configureRequest, deleteCookie } from '~/helpers/gIPTools'
+import frApi from '~/services/fuelrats'
 
 import actionTypes from '../actionTypes'
 import {
+  selectSessionToken,
   selectPageRequiresAuth,
   selectSession,
   selectUserById,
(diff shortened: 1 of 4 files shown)


UncleClapton (changes requested):
Hey! Please add changelog entry for this :)

src/store/actions/session.js
```diff
@@ -22,9 +24,22 @@ import { getUserProfile } from './user'
  * @returns {Function} Redux action thunk
  */
 export const logout = (ctx) => {
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
     deleteCookie('access_token', ctx)
 
+    const token = selectSessionToken(getState())
+
+    await dispatch(createAxiosFSA(
```
UncleClapton: so I know I did this, but awaiting this without any UI feedback is probably bound to cause some confusion. I think it's easiest to just remove the await and do this step in the background without the user knowing it.

UncleClapton (changes requested):

src/store/actions/session.js
```diff
@@ -22,9 +24,22 @@ import { getUserProfile } from './user'
  * @returns {Function} Redux action thunk
  */
 export const logout = (ctx) => {
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
     deleteCookie('access_token', ctx)
 
+    const token = selectSessionToken(getState())
```
UncleClapton: There's a second call to `getState()` in this same function now, Could you extract that to a separate variable?

UncleClapton (changes requested):

UncleClapton (approved):

PR #332, opened 2021-07-10
Feat/web 224/profile pic uploader

Pero-Moretti:
Now de-linted and resolved initial PR comments.


src/components/ProfileUserAvatar/ProfileUserAvatar.js
```diff
@@ -0,0 +1,100 @@
...
+
+    return (
+      <>
+        <div className="user-avatar" onClick={this._handleToggleUploadAvatar}>
+          <div className="avatar xl">
+            <Image
+              unoptimized
+              alt="User's avatar"
+              height={170}
+              src={`${appUrl}${userAvatar}`}
```
UncleClapton: relying on `appUrl` here is no longer required. We can feed the avatar path directly into `<Image />`

src/components/ProfileUserAvatar/ProfileUserAvatar.js
```diff
@@ -0,0 +1,100 @@
...
+  selectCurrentUserHasScope,
+} from '~/store/selectors'
+
+import UploadAvatarModal from '../UploadAvatarModal'
+
+const { publicRuntimeConfig } = getConfig()
+const { appUrl } = publicRuntimeConfig
+
+@connect
+class ProfileHeader extends React.Component {
```
UncleClapton: This whole class could be rewritten as a Function component with Hooks.

src/store/actions/user.js
```diff
@@ -80,3 +78,18 @@ export const updateUser = (data, password) => {
...
 }
+
+export const updateAvatar = (user, datamime, data) => {
+  return (dispatch, getState) => {
+    const formData = new FormData();
+    formData.append('image', data)
+    const request = {
+      url: `/users/${user}/image`,
+      method: 'post',
+      contenttype: datamime,
```
UncleClapton: content type cannot be set like this, and since this is formData it doesn't matter anyway. There's no need to really care about the mime type of the input image

src/scss/components/_upload-avatar-modal.scss
```diff
@@ -0,0 +1,29 @@
+.zoomandcrop {
```
UncleClapton: CSS classes should be camelCase, and new components should use CSS Modules instead of being added to the global stylesheet.

This ones on me for not warning about this. 

If you're new to the concept, A good example of using css modules would be something like the password fieldset under `src/components/Fieldsets/PasswordFieldset`

UncleClapton (changes requested):
these are the biggest things that stand out to me. I'm going to hold off on fine-tooth combing it until after you've linted and cleaned up your code.

Take your time, there's no rush on this 😄

src/components/UploadAvatarModal/UploadAvatarModal.js
```diff
@@ -0,0 +1,247 @@
...
+              aria-labelledby="Zoom"
+              onChange={(zoom) => { setZoom(zoom) }}
+            />
+          </div>
+          <div class='zoomslidericon'>
+            <FontAwesomeIcon icon="search-minus" />
+          </div>
+        </div>
+        <div class="rotateandcrop">
+          <div className="crop-container">
```
UncleClapton: One thing I noticed from your preview is that the copper is at an odd aspect ratio, with black bars on either side. was that a side effect of the page size or is it always like that? If it's always like that, we should probably see if it's possible to ensure it's always squared up unless page height/width restrictions require us to shrink it.

Pero-Moretti:
I've fixed the comments and delinted. Only one thing I'm not 100% happy about, which is a render glitch on profile page load. Very briefly shows you the upload button. See what you think.

src/components/UploadAvatarModal/UploadAvatarModal.module.scss
```diff
@@ -0,0 +1,37 @@
+  .zoomAndCrop {
```
UncleClapton: We have Lint rules for CSS as well 😄

src/components/UploadAvatarModal/UploadAvatarModal.js
```diff
@@ -0,0 +1,247 @@
...
+  // Linting note: Cannot use arrow function in onevent, need to wrap
+  const handleSetCrop = useCallback((newCrop) => {
+    setCrop(newCrop)
+  }, [])
+  const handleSetZoom = useCallback((newZoom) => {
+    setZoom(newZoom)
+  }, [])
+  const handleSetRotation = useCallback((newRotation) => {
+    setRotation(newRotation)
+  }, [])
```
UncleClapton: This is a useless wrapping of `setState` funcs. All of these callbacks could be bound directly to the event listener props. for instance:

```jsx
const [crop, handleCropChange] = useState({ x: 0, y: 0 })

/* ... */

return (<Cropper onCropChange={handleCropChange} />)
```

src/components/ProfileUserAvatar/ProfileUserAvatar.js
```diff
@@ -0,0 +1,45 @@
...
+import styles from './ProfileUserAvatar.module.scss'
+
+function ProfileUserAvatar () {
+  const userAvatar = useSelectorWithProps({ size: 170 }, withCurrentUserId(selectAvatarByUserId))
+
+  const [showUploadAvatar, setShowUploadAvatar] = useState(false)
+
+  const handleToggleUploadAvatar = useCallback(() => {
+    setShowUploadAvatar(!showUploadAvatar)
+  }, [showUploadAvatar])
```
UncleClapton: `useState()` setters can accept a function which accepts the current value of the hook. Doing this lets us remove the `useCallback()` dependency on the `useState()` value here. 

```suggestion
  const handleToggleUploadAvatar = useCallback(() => {
    setShowUploadAvatar((prevState) => {
      return !prevState
    })
  }, [])
```

src/components/ProfileUserAvatar/ProfileUserAvatar.js
```diff
@@ -0,0 +1,45 @@
...
+      <div className={styles.userAvatar}>
+        <div className="avatar xl">
+          <Image
+            unoptimized
+            alt="User's avatar"
+            height={170}
+            src={userAvatar}
+            width={170} />
+        </div>
+        <div className={styles.userAvatarEdit}>
```
UncleClapton, line 33: I'm not exactly sure why that graphical bug exists here, but I will pull down the PR locally and load it up to see. Will update this with a change suggestion when I get it working!

src/components/UploadAvatarModal/UploadAvatarModal.js
```diff
@@ -0,0 +1,247 @@
...
+function UploadAvatarModal (props) {
+  const {
+    onClose,
+    isOpen,
+  } = props
+
+  const [result, setResult] = useState({})
+
+  const [upImg, setUpImg] = useState()
+  const [crop, setCrop] = useState({ x: 0, y: 0 }) // Lint comment on "Identifier name is too short" - this is required by the Cropper. Cannot change.
```
UncleClapton: Please do not leave lint comments like this. Linting is enforced by our CI, and will refuse to deploy when left unresolved. if something like this MUST remain (in this case it does), then use a `eslint-disable-next-line` directive for the reporting rule.

```suggestion
  // eslint-disable-next-line id-length -- Required by react-easy-crop
  const [crop, setCrop] = useState({ x: 0, y: 0 })
```

src/components/UploadAvatarModal/UploadAvatarModal.js
```diff
@@ -0,0 +1,247 @@
...
+  const [zoom, setZoom] = useState(1)
+  const [rotation, setRotation] = useState(0)
+  const [croppedAreaPixels, setCroppedAreaPixels] = useState(null)
+  const [submitReady, setSubmitReady] = useState(false)
+  const [submitting, setSubmitting] = useState(false)
+
+  // Linting note: Cannot use arrow function in onevent, need to wrap
```
UncleClapton: Same as above, Please do not leave "lint notes" like this. The linter makes this pattern of requiring stable functions consistent across the project, therefore there's little value in pointing that out here.

src/components/UploadAvatarModal/UploadAvatarModal.js
```diff
@@ -0,0 +1,247 @@
...
+      })
+    }
+  }, [croppedAreaPixels, rotation, submit, upImg])
+
+  return (
+    <ModalContent className="dialog no-pad">
+      <UploadAvatarMessageBox result={result} />
+      <div>
+        <label>{'Select Image: '}</label>
+        <input accept="image/*" aria-label="Select Image" type="file" onChange={onSelectFile} />
```
UncleClapton: Would be great if we could limit this to just the file types we accept. See ticket for supported image types. all of their MIMES are pretty straightforward.

UncleClapton (changes requested):
Overall pretty good stuff. For the future, I would recommend looking into integrating ESLint and StyleLint into your IDE/Editor. You'll get real-time feedback and often times automatic fixes / suggestions. There's extensions for each on most modern code editors.

I'm still looking into that CSS issue with the modal opener button. I'll let you know when I find out why. If it's simple enough I'll just apply the fix along with the merge to develop.

src/components/UploadAvatarModal/UploadAvatarModal.js
Pero-Moretti: Umm... I tried that to start with, but lint complained bitterly that I wasn't allowed... I 100% agree it's useless otherwise...

Pero-Moretti: This is very frustrating - I'd not wrapped the functions previously, I believe exactly the way you demonstrated, but something was broken - I don't recall exactly what the issue was, either lint complained bitterly or the sliders simply stopped working, so I ended up wrapping as you see. However, changing them back just now everything is working. I have no idea why. Because I 100% agree with you that such a wrap was useless otherwise.

Oh well.

src/components/UploadAvatarModal/UploadAvatarModal.js
Pero-Moretti: As far as I'm aware, all images ultimately uploaded to the server are *going* to be png, because of the cropper tool drawing whatever image to a hidden canvas for the transforms, then pulling the data of the canvas off (which comes off as png). Is there a need to restrict what the user can do in this case? Or should we just check that the canvas data being pulled off is one of the three types?

src/components/UploadAvatarModal/UploadAvatarModal.module.scss
Pero-Moretti: On it...

src/components/UploadAvatarModal/UploadAvatarModal.js
UncleClapton: Uhh that's a fair point. Could we test for GIFs to see how they get processed? I assume it'll just take the first frame but I wanna be sure.

src/components/UploadAvatarModal/UploadAvatarModal.js
```diff
@@ -128,7 +119,16 @@ function UploadAvatarModal (props) {
...
+            let start = 0
+            let mimeend = 0
+            let datastart = 0
+            if (reader.result.indexOf('data:') === 0) {
+              start = TRIM_DATA
+            }
+            mimeend = reader.result.indexOf(';', start)
+            datastart = reader.result.indexOf(',', mimeend)
+            const mime = reader.result.substring(start, mimeend)
+            const b64data = reader.result.substr(datastart + 1)
```
UncleClapton: Since we're always outputting as PNG, why not just do something like:

```js
const [,b64data] = reader.result.split(',')
```

and then we just define a constant for the mime.

UncleClapton:
@Pero-Moretti Just those couple things there about your image processing code and I think we're all good.

src/components/UploadAvatarModal/UploadAvatarModal.js
Pero-Moretti: I wasn't confident that'd be the case, but since we don't actually use the mime type anyway no harm in that. I'll push again tomorrow.

UncleClapton:
Just leaving this note here. I'm going to merge this PR once the final change is in. Any further changes/fixes (mainly that graphical bug) should be made in a follow-up

src/components/UploadAvatarModal/UploadAvatarModal.js
Pero-Moretti: I'd not thought of animated gifs. Because the cropper is a standard  it'll animate it while editing, but when you click "upload" it does indeed only show the first frame. It gives the impression your cropped avatar will animate, which it won't. I think that's a good enough reason not to have image/* :(

Pero-Moretti:
All done.

UncleClapton:
> It works, but not in the way the user expects.

Yeah probably best to forbid it just for the sake of leaving them with the expectation that animated images are not supported.

UncleClapton (approved):
LGTM!

PR #417, opened 2025-01-05
Fix typo on history page

evertiro:
Just after the new pages went live, it was pointed out to me that we use the spelling "knowledge base" on Confluence, but I had used "knowledgebase" on the site. This tiny update fixes that oversight to keep things consistent.


UncleClapton (approved):
LGTM!

PR #418, opened 2025-01-07
Redesign platform indicators on dispatch board

### LINKED ISSUE #413: Change style of platform indicators in dispatch board
### Description

Change the style of the platform indicators in the dispatch board to use a coloured square with text inside much like the game mode indicators 

_No response_

_No response_

### I can help!

- [ ] I am willing to implement this feature, and submit it as a PR.

evertiro:
Rework platform indicators into a cleaner, simpler design. Note that there is a preexisting error with linting that has not been resolved due to the ps and xb classes being too short per our rules. Wasn't sure how to deal with this and since it's preexisting, I chose to leave it alone for now.

Fixes #413

### DIFF
--- a/src/components/DispatchTable/DispatchTable.module.scss
+++ b/src/components/DispatchTable/DispatchTable.module.scss
@@ -93,26 +93,6 @@
 
     text-align: center;
 
-    &.horizons3 > .platformBadgeLabel {
-      background-color: $magenta-dark3;
-    }
-
-    &.horizons4 > .platformBadgeLabel {
-      background-color: $blue-dark3;
-    }
-
-    &.odyssey > .platformBadgeLabel {
-      background-color: $gold-dark2;
-    }
-
-    &.ps > .platformBadgeLabel {
-      background-color: $xbox-green-light;
-    }
-
-    &.xb > .platformBadgeLabel {
-      background-color: $playstation-blue-light3;
-    }
-
     > .platformBadgeIcon {
       display: inline-block;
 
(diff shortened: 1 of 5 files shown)


src/components/DispatchTable/RescueRow.js
```diff
@@ -113,19 +113,44 @@ function RescueRow (props) {
...
-            )
-          }
         </CopyToClipboard>
       </td>
       <td
         className={['rescue-row-platform', styles.platform, styles[rescue.attributes.platform]]}
         title={rescuePlatform.long}>
-        {rescuePlatform.short}
+        {
+          platform === 'pc' && expansion && (
```
UncleClapton, line 122: if by chance we don't get an expansion, we'd end up with no platform displayed with this logic. Can we get a fallback case here? Maybe just by defaulting to "PC" as a label.

src/components/DispatchTable/DispatchTable.module.scss
```diff
@@ -75,46 +75,61 @@
...
+    &.ps > .platformBadgeLabel {
+      background-color: $xbox-green-light;
+    }
 
-  &.xb {
-    color: $xbox-green-light
-  }
+    &.xb > .platformBadgeLabel {
+      background-color: $playstation-blue-light3;
+    }
```
UncleClapton: We should avoid useless hoisting of modifier classes like this. I'd say it would be better for these to exist as modifiers within the `platformBadgeLabel` class instead


Ideally we'd also try to keep our style structure within CSS modules as flat a possible, as they all complile to scoped CSS names anyway. There's no need to hold a definite element structure when the compiler assures all resulting class names will be unique. I believe this file is especially bad in that regard simply because it was faster to copy&paste the old stylesheet instead of recreating it for modules 😂. No need to do a big cleanup right now if you don't want to, but just wanted to call it out cuz it's easy to paint yourself into a corner by strictly defining your styles in this way.

UncleClapton (changes requested):
just got a couple things here

src/components/DispatchTable/DispatchTable.module.scss
UncleClapton: @evertiro this was left unaddressed 🙂

UncleClapton (approved):
LGTM, Awesome work :)

PR #419, opened 2025-01-08
Quick pass at adding line numbers to dispatch table quotes

### LINKED ISSUE #412: Dispatch line numbers for quotes
### Description

Display line numbers for quotes that corresponds to the index for !sub to make substitutions easier

_No response_

SuperManifolds

### I can help!

- [ ] I am willing to implement this feature, and submit it as a PR.

evertiro:
Pretty much what the title says. Not sure this is the best way to do it, but it's the first option I thought of that works well - at least as long as we have less than 100 open cases.

Fixes #412


UncleClapton (approved):
LGTM!

PR #421, opened 2025-03-16
Remove disable profile button from profile

evertiro:
Since the disable profile option has been removed from the API, let's remove the button that doesn't work too. I've left the modal alone for now, as I still think we should leave the functionality alone and just remove access for the user to disable their own account. That said, if the consensus is that the functionality should be removed completely, it's easy enough to clean up.

### DIFF
--- a/src/components/ProfileHeader/ProfileHeader.js
+++ b/src/components/ProfileHeader/ProfileHeader.js
@@ -13,7 +13,6 @@ import formatAsEliteDateTime from '~/util/date/formatAsEliteDateTime'
 
 import ChangeEmailModal from '../ChangeEmailModal'
 import ChangePasswordModal from '../ChangePasswordModal'
-import DisableProfileModal from '../DisableProfileModal'
 import ProfileUserAvatar from '../ProfileUserAvatar'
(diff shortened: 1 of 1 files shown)


SuperManifolds (dismissed):
Leaving the functionality and just removing the button sounds good to me

UncleClapton:
We can always refer to git history if we need to bring dead code back. I feel we should remove it in it's entirety if the feature isn't even supported on the API anymore.

At the very least lets remove the modal component and related state code from the page so we're no longer bundling dead code.

UncleClapton (changes requested):
see above. forgor to put my comment as a review :P

evertiro:
> We can always refer to git history if we need to bring dead code back. I feel we should remove it in it's entirety if the feature isn't even supported on the API anymore.
> 
> At the very least lets remove the modal component and related state code from the page so we're no longer bundling dead code.

My thought was that the change to the API should be reverted. I get why the change was made (spur of the moment change to prevent further headaches), but it really should have been done on the UI side.

UncleClapton (approved):
Looks removed to me :)

we hashed this out over a call. Leaving modal code in, removed remaining references from live code so it's no longer bundled

PR #428, opened 2025-03-19
Actually copy commander name on CMDR col

### LINKED ISSUE #427: Dispatch board copies wrong data for CMDR column
From a discussion on IRC: Currently, clicking the copy button on the CMDR column copies the IRC nick instead of the actual commander name. The biggest benefit from having that field copyable is for quick lookups to check online status. We also have a separate field in the detail view for IRC nick, which is also copyable. The CMDR column should really copy the data its actually presenting, not another field entirely.

evertiro:
Title says what it does, copies commander name instead of IRC nick when clicking the CMDR column. Fixes #427

### DIFF
--- a/src/components/DispatchTable/RescueRow.js
+++ b/src/components/DispatchTable/RescueRow.js
@@ -79,7 +79,6 @@ function RescueRow (props) {
   const {
     codeRed,
     status,
-    clientNick,
     client,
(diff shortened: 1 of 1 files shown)


UncleClapton (approved):
LGTM!

Issue #411, opened 2024-11-05
Display system information in dispatch board


SuperManifolds:
### Description

Show system information in dispatch board (distance from landmark, star type)

### FuelRats IRC Nickname (optional)

SuperManifolds

### I can help!

- [ ] I am willing to implement this feature, and submit it as a PR.

evertiro:
Thoughts on how you'd like to see this displayed? We don't have a ton of space to work with given the current design.

UncleClapton:
TODO: Get SAPI Documentation for this

UncleClapton:
@evertiro this is a request to return a feature we cut at some point due to issues, though I can't remember the specifics why.

We used to display some basic system information in chips on the detail view, similar to the paperwork link chip

evertiro:
I'm good with issues...

SuperManifolds:
We could go two ways with this, since I implemented this in mecha already I could write up some documentation on how to achieve this, though mecha actually already writes a cache of system information inside the rescue entry to avoid repeated lookups, so we could potentially just use that

kenneaal:
SAPI's JSONSchema is at https://systems.api.fuelrats.com/metadata/JSONSchema
That should provide the field info needed to unpack results from queries; given that we at least will have systemID64, everything can be gotten fast from https://systems.api.fuelrats.com/api/[system|body]?id=<id> queries.

evertiro:
> SAPI's JSONSchema is at https://systems.api.fuelrats.com/metadata/JSONSchema That should provide the field info needed to unpack results from queries; given that we at least will have systemID64, everything can be gotten fast from [https://systems.api.fuelrats.com/api/[system|body]?id=](https://systems.api.fuelrats.com/api/%5Bsystem%7Cbody%5D?id=) queries.

Appreciate the pointer!

evertiro:
Resolved with yesterdays release

PR #431, opened 2025-03-21
Display system info in dispatch board

evertiro:
I'm not confident I'm done with this, but it does work as is and covers the things listed in #411, so it's probably worth getting eyes on it.


UncleClapton (approved):
LGTM! Nice work :)

PR #432, opened 2025-04-18
First pass at getting rid of react-table

evertiro:
This is a work-in-progress build intended to replace the react-table dependency with a custom solution abstracted from the pagination component currently used by the blog.

Current changes:

- [x] Blog pagination component reworked into a reusable pagination component
- [x] Leaderboard rewritten to use an ordered list and the new pagination component
- [x] Pagination component needs to support a page input field
- [x] Pagination component needs to support setting number of entries per page


evertiro:
This is mostly working, but parts of it are definitely a bit hacky. The one thing that _isn't_ currently working is multi-page filtering. Problem is, I hate how I'm currently handling query building for the limit and filter params, and want to come up with a better way of handling it before putting more effort into the current iteration.

src/components/Leaderboard/RescueAchievementIcon.js
```diff
@@ -42,7 +45,7 @@ function RescueAchievementIcon ({ rescueCount, ...iconProps }) {
 
   return Icon && (
     <div
-      className={['achievement rescue-count', { crown: count >= 500 }]}
+      className={classes}
```
UncleClapton: This can still be written similarly to before like this:

```suggestion
      className={[`${styles.achievement} ${styles.rescueCount}`, { [styles.crown]: count >=500 }]}
```

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,133 @@
...
+
+  if (pageSize !== DEFAULT_PAGE_SIZE) {
+    prevPage = makePaginatedRoute({
+      route, author, category, page: Math.max(1, page - 1), limit: pageSize,
+    })
+
+    nextPage = makePaginatedRoute({
+      route, author, category, page: Math.min(page + 1, totalPages), limit: pageSize,
+    })
+  }
```
UncleClapton, line 47: Instead of accepting `route`, `author`, and `category` as props, I think it would be better to pass the control of route formation off to the parent. Instead I'd recommend a route generator prop similar to:

```tsx
const makeLeaderboardRoute = (params) => makeRoute('leaderboard', params)

function Leaderboard () {
  return (
    <Pagination
      onMakeNextRoute={makeLeaderboardRoute} />
  )
}
```

This has a few benefits. Not only does it simplify the props surface, it also lets the pagination component conform to the route definition of the page, rather than requiring a strict page structure. This also means that the only thing `<Pagination />` needs to worry about is calculating the page variable, as well as passing along the pageSize (limit) prop, but perhaps `pageSize` state could be internalized?

src/util/router/makePaginatedRoute.js
```diff
@@ -2,30 +2,32 @@ import makeRoute from './makeRoute'
...
 /**
  * @param {object} params
+ * @param {string} params.route
  * @param {string|number|null} params.author
  * @param {string|number|null} params.category
  * @param {string|number|null} params.page
  * @returns {string}
  */
-export default function makeBlogRoute (params) {
+export default function makePaginatedRoute (params) {
```
UncleClapton, line 11: If we refactor to `onMakeNextRoute`, we probably can revert this to the original `makeBlogRoute`.

Because short links to these routes exist and must be maintained, this route structure for blog posts are set in stone.

New routes like what we're doing here don't need to follow this strictly, however, so we could just use a query param for pagination on leaderboard

UncleClapton (changes requested):
Overall it's a step in a good direction, though I definitely see what you mean by hacky. hopefully the suggestions I've got help some of that.

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,133 @@
...
+      {
+        (onUpdatePageSize) && (
+          <div className={styles.perPage}>
+            <select aria-label="Per Page" value={pageSize} onChange={onUpdatePageSize}>
+              <option value="5">{'5 Rows'}</option>
+              <option value="10">{'10 Rows'}</option>
+              <option value="20">{'20 Rows'}</option>
+              <option value="25">{'25 Rows'}</option>
+              <option value="50">{'50 Rows'}</option>
+              <option value="100">{'100 Rows'}</option>
```
UncleClapton, line 105: Future nice to have:

Abstract this out. Page row options should be props-controlled.

```tsx
const pageSizes = [5, 10, 20, 25, 50, 100]

// ...
function Leaderboard () {
  return (
    <Pagination
      pageSize={pageSizeState}
      pageSizes={pageSizes}
      // This feels wrong cuz we have both `pageSize` and `defaultPageSize`. 
      // typically a no-no for input components, and technically `pageSize`
      // maps to an input...
      defaultPageSize={pageSizes[3]} 
      onUpdatePageSize={handleUpdatePageSize} />
  )
}
```

Side note: 

Its easy to forget how much of a pain PropTypes is sometimes compared to TS / Static typing. I started thinking about how you'd conditionally require `defaultPageSize`, and compared to TS, it's kinda a pain and not very intuitive to read.

PR #433, opened 2025-04-24
Temporary fix for merch store link

evertiro:
It was the first thing I asked when I joined, we just had another ticket on it, and it's asked regularly in ratchat. Plus, having a broken link is bad practice. And I have goals that will necessitate the page anyway.

### DIFF
--- a/src/pages/store/index.js
+++ b/src/pages/store/index.js
@@ -20,7 +20,7 @@ function Store () {
 Store.getPageMeta = () => {
   return {
     title: 'Merch Store',
-    description: 'Support the Fuel Rats\' mission to aid spaceship pilots in our vast universe! Your generous donation helps us continue our in-game rescue services, ensuring no player is left stranded in space. Contribute today and fuel our journey to make the gaming experience enjoyable for everyone!',
+    description: 'Support the Fuel Rats\' mission to aid spaceship pilots in our vast universe!',
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/pages/store/index.js
```diff
@@ -0,0 +1,31 @@
...
+        <Link href="/donate">{'donations'}</Link>{'.'}
+      </div>
+    </div>
+  )
+}
+
+Store.getPageMeta = () => {
+  return {
+    title: 'Merch Store',
+    description: 'Support the Fuel Rats\' mission to aid spaceship pilots in our vast universe! Your generous donation helps us continue our in-game rescue services, ensuring no player is left stranded in space. Contribute today and fuel our journey to make the gaming experience enjoyable for everyone!',
```
UncleClapton: in interest of maintaining tech debt, could we at least resolve this line length issue? the others we can handle eventually, or you could resolve them all in this PR if you'd like, but at the very least lets not introduce more work

UncleClapton (changes requested):

src/pages/store/index.js
evertiro: I have no issue with that, I just copied the meta description from the donation page. Good practice would be to have meta descriptions be ~105 chars in length max anyway.

evertiro:
For now, just removed the donation-specific stuff and redundant wording. We'll have to come up with a more appropriate meta description for long-term use eventually.

UncleClapton (approved):
LGTM!

PR #434, opened 2025-04-27
Fix lint issues

evertiro:
This fixes all current lint issues except for the selector-class-pattern errors on the .ps and .xb classes. I'm not sure the best way to address those two issues since the issue is their length, and changing those selectors will impact other things due to how they're generated.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,19 +2,19 @@
   "[javascript]": {
     "editor.defaultFormatter": "dbaeumer.vscode-eslint",
     "editor.codeActionsOnSave": {
-      "source.fixAll.eslint": true,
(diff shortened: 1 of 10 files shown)


evertiro:
Note: We could cheat, and simply disable the rule for those two lines by adding `/* stylelint-disable-line selector-class-pattern */`

UncleClapton:
you're a damn psychic, you know that? I was gonna do this tonight 🤣 

Could you just stick a stylelint-disable directive on the platform classnames, with a comment citing that it's to align with the platform values coming from the api?

evertiro:
Your wish, my command!

UncleClapton (approved):
lgtm!

PR #449, opened 2025-07-31
feat(UserNicknamesPanel): add ability to set IRC display nickname

### LINKED ISSUE #448: As a rat I want to be able to change my IRC display nick from the profile
### Description

Allow rats to change their IRC display nick from the fuelrats.com profile so they don't have to navigate the confusing NickServ SET options
Acceptance Criteria
* There is a button to set a new nick as the display nick
* The button has a tooltip that explains what action it performs
* The UI updates automatically to reflect the new changes when a new display nick is set

_No response_

_No response_

### I can help!

- [x] I am willing to implement this feature, and submit it as a PR.

SuperManifolds:
Closes #448 

## Description

- Add setDisplayNickname action to call PUT /nicknames/:id/display endpoint
- Add star icon button to set any nickname as display nickname
- Update UI to show star/delete buttons only for non-display nicknames
- Add faStar icon to fontawesome library
- Refresh user profile after setting display nickname to update UI

### Additional Information

*

### Demo

https://fuelrats.cloud/s/4NiNcdZXgfTLFsH

### Contributor's Checklist

* [x] This PR was created to resolve an existing issue or set of issues.
* [x] This PR satisfies any and all acceptance criteria laid out by issue(s) it resolves.
* [x] I have discussed creating this PR with the maintainers in the issue(s) beforehand.
* [x] I have thoroughly tested the changes this PR introduces in a local development environment.
* [x] I have linted the entire codebase using `yarn lint` and confirmed there are no errors or warnings.
* [x] I have followed the commit conventions laid out by [`CONTRIBUTING.md`](https://github.com/FuelRats/fuelrats.com/blob/develop/CONTRIBUTING.md#commit-conventions).

### Maintainer's Checklist

* [ ] User-facing changes this PR introduces have been captured in a changeset.
* [ ] This PR has been linted and tested locally as a part of the review process.
* [ ] All issues this PR resolves have been properly linked before merging.

### DIFF
--- a/.changeset/irc-display-nickname.md
+++ b/.changeset/irc-display-nickname.md
@@ -1,0 +1,9 @@
+---
+"fuelrats.com": minor
+---
+
+Add ability to set IRC display nickname in user profile
+
+- Users can now select which of their registered IRC nicknames should be their display nickname
+- Added a star icon button next to each non-display nickname to set it as the display nickname
+- The current display nickname is indicated by the absence of action buttons
--- a/.changeset/irc-display-nickname.md
+++ b/.changeset/irc-display-nickname.md
@@ -4,6 +4,8 @@
 
 Add ability to set IRC display nickname in user profile
 
-- Users can now select which of their registered IRC nicknames should be their display nickname
-- Added a star icon button next to each non-display nickname to set it as the display nickname
-- The current display nickname is indicated by the absence of action buttons
+Users can now choose which of their registered IRC nicknames appears as their primary display name. In the IRC Nicknames section of your profile:
+
+- A filled star indicates your current display nickname
+- Click the outline star next to any other nickname to set it as your new display nickname
+- Hover over the star icons for helpful tooltips
--- a/src/components/UserNicknamesPanel/UserNicknamesPanel.js
+++ b/src/components/UserNicknamesPanel/UserNicknamesPanel.js
@@ -99,36 +99,46 @@ function UserNicknamesPanel () {
               const isDisplayNick = nickname.attributes?.display === nickname.attributes?.nick
               return (
                 <li key={nickname.id}>
-                  <span>{nickname.attributes?.nick}</span>
-                  <div>
+                  <span>
+                    {nickname.attributes?.nick}
                     {
-                      // Only show set display button for non-display nicks
-                      !isDisplayNick && (
-                        <ConfirmActionButton
-                          className="icon"
-                          confirmButtonText={`Set '${nickname.attributes?.nick}' as display nickname`}
-                          confirmSubText=""
-                          denyButtonText="Cancel"
-                          name={nickname.id}
-                          onConfirm={handleSetDisplayNickname}
-                          onConfirmText="">
-                          <FontAwesomeIcon fixedWidth icon="star" />
-                        </ConfirmActionButton>
+                      isDisplayNick && (
+                        <FontAwesomeIcon
+                          className={styles.displayIcon}
+                          fixedWidth
+                          icon="star"
+                          title="Current display nickname" />
                       )
                     }
+                  </span>
+                  <div className={styles.controlContainer}>
                     {
-                      // Only render for additional nicks, prevent for display nick.
+                      // Only show buttons for non-display nicks
                       !isDisplayNick && (
-                        <ConfirmActionButton
-                          className="icon"
-                          confirmButtonText={`Delete nickname '${nickname.attributes?.nick}'`}
-                          confirmSubText=""
-                          denyButtonText="Cancel"
-                          name={nickname.id}
-                          onConfirm={handleDeleteNickname}
-                          onConfirmText="">
-                          <FontAwesomeIcon fixedWidth icon="trash" />
-                        </ConfirmActionButton>
+                        <>
+                          <ConfirmActionButton
+                            className="icon"
+                            confirmButtonText={`Set '${nickname.attributes?.nick}' as display nickname`}
+                            confirmSubText=""
+                            denyButtonText="Cancel"
+                            name={nickname.id}
+                            onConfirm={handleSetDisplayNickname}
+                            onConfirmText=""
+                            title="Set as display nickname">
+                            <FontAwesomeIcon fixedWidth icon={['far', 'star']} />
+                          </ConfirmActionButton>
+                          <ConfirmActionButton
+                            className="icon"
+                            confirmButtonText={`Delete nickname '${nickname.attributes?.nick}'`}
+                            confirmSubText=""
+                            denyButtonText="Cancel"
+                            name={nickname.id}
+                            onConfirm={handleDeleteNickname}
+                            onConfirmText=""
+                            title="Delete nickname">
+                            <FontAwesomeIcon fixedWidth icon="trash" />
+                          </ConfirmActionButton>
+                        </>
                       )
                     }
                   </div>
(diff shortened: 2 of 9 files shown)


src/components/UserNicknamesPanel/UserNicknamesPanel.js
```diff
@@ -83,24 +96,42 @@ function UserNicknamesPanel () {
...
+                          confirmButtonText={`Delete nickname '${nickname.attributes?.nick}'`}
+                          confirmSubText=""
+                          denyButtonText="Cancel"
+                          name={nickname.id}
+                          onConfirm={handleDeleteNickname}
+                          onConfirmText="">
+                          <FontAwesomeIcon fixedWidth icon="trash" />
+                        </ConfirmActionButton>
+                      )
+                    }
```
UncleClapton, line 139: since this is the same condition, I think it would be clearer to express it as a single conditional


```tsx
                    {
                      // Only show buttons for non-display nicks
                      !isDisplayNick && (
                        <>
                          <ConfirmActionButton
                            className="icon"
                            confirmButtonText={`Set '${nickname.attributes?.nick}' as display nickname`}
                            confirmSubText=""
                            denyButtonText="Cancel"
                            name={nickname.id}
                            onConfirm={handleSetDisplayNickname}
                            onConfirmText="">
                            <FontAwesomeIcon fixedWidth icon="star" />
                          </ConfirmActionButton>
                          <ConfirmActionButton
                            className="icon"
                            confirmButtonText={`Delete nickname '${nickname.attributes?.nick}'`}
                            confirmSubText=""
                            denyButtonText="Cancel"
                            name={nickname.id}
                            onConfirm={handleDeleteNickname}
                            onConfirmText="">
                            <FontAwesomeIcon fixedWidth icon="trash" />
                          </ConfirmActionButton>
                        </>
                      )
                    }
```

src/components/UserNicknamesPanel/UserNicknamesPanel.module.scss
```diff
@@ -21,6 +21,11 @@
   span {
     word-break: break-word;
   }
+
+  div {
+    display: flex;
+    gap: 0.5rem;
+  }
```
UncleClapton: we should avoid adding further structure-bound styles and instead define this as it's own style at the top level of the file. maybe something like `.controlContainer`?

UncleClapton (changes requested):

.changeset/irc-display-nickname.md
```diff
@@ -0,0 +1,9 @@
+---
+"fuelrats.com": minor
+---
+
+Add ability to set IRC display nickname in user profile
+
+- Users can now select which of their registered IRC nicknames should be their display nickname
+- Added a star icon button next to each non-display nickname to set it as the display nickname
+- The current display nickname is indicated by the absence of action buttons
```
UncleClapton: Thinking on this more, I think this could be potentially confusing to a newcomer.

What if we used a filled in star in place of the buttons on the display name, and then the "set display" button uses an outline star? https://fontawesome.com/v5/icons/star?f=classic&s=regular

UncleClapton: perhaps we also give that "display" star some description on hover to explain what it means

src/util/fontawesome/library.js
```diff
@@ -36,6 +36,7 @@ export {
   faSignature,
   faSpaceShuttle,
   faSpinner,
+  faStar,
```
UncleClapton, line 39: we also need to provide the regular version, since this only provides the solid version. Sorry I forgot we don't actually have this icon pack in here yet. without adding it the icon won't load properly.

Screenshot 2025-09-07 at 11 01 08


To fix:

1. run `yarn add "@fortawesome/free-regular-svg-icons@^5.15.4"`
2. add export to `library.js`

```ts
export {
  faStar as farStar,
} from '@fortawesome/free-regular-svg-icons'
```

src/components/UserNicknamesPanel/UserNicknamesPanel.js
```diff
@@ -57,6 +57,19 @@ function UserNicknamesPanel () {
...
+    const nickname = nicknames.find((nick) => {
+      return nick.id === event.target.name
+    })
+
+    const response = await dispatch(setDisplayNickname(nickname.id, nickname.attributes.nick))
+
+    if (!isError(response)) {
+      // Refresh user profile to get updated nicknames
+      await dispatch(getUserProfile())
+    }
```
UncleClapton, line 62: Sorry I forgot about this, but because the way I handled loading states with confirmation buttons is weird, this function always needs to `return true`, or at least some truthy value. Otherwise we get an infinite spinner.

src/components/UserNicknamesPanel/UserNicknamesPanel.js
```diff
@@ -57,6 +57,19 @@ function UserNicknamesPanel () {
     return undefined
```
UncleClapton: this is unrelated to your changes, but could you add `return true` here too?

src/components/UserNicknamesPanel/UserNicknamesPanel.js
```diff
@@ -57,6 +57,19 @@ function UserNicknamesPanel () {
...
   }, [dispatch, nicknames, user])
 
+  const handleSetDisplayNickname = useCallback(async (event) => {
+    const nickname = nicknames.find((nick) => {
+      return nick.id === event.target.name
+    })
+
+    const response = await dispatch(setDisplayNickname(nickname.id, nickname.attributes.nick))
+
+    if (!isError(response)) {
```
UncleClapton: We could file this as a future enhancement instead, but I think we could easily add error handling  here. We already show errors for deletion, so we can re-use the `error` `useState` for that without changing the UI.

If we do change this now, could we also add `setError(null)` at the top of both action handlers? That way we reset the error state making it clear that if the error repeats, it's the error for the latest user interaction, and not a previous one.

.changeset/irc-display-nickname.md
UncleClapton: Could we update this to reflect the final UX?

UncleClapton (changes requested):
Hey Ran into a 500 from the API when setting the display nick. Would that just be because we're in lowers?

error id: `71de695d-0143-48c6-ba46-ca2608529495`

Glad it did error though, because I entirely forgot controlling the loading state for action buttons is wonky.

SuperManifolds:
I suspect the error 500 is due to funkiness with IRC server interaction on the dev API, I had already tested this locally and it works, I will look into why it's not working on the dev api

src/components/UserNicknamesPanel/UserNicknamesPanel.js
```diff
@@ -125,7 +121,7 @@ function UserNicknamesPanel () {
                             title="Set as display nickname"
                             onConfirm={handleSetDisplayNickname}
                             onConfirmText="">
-                            <FontAwesomeIcon fixedWidth icon={['far', 'star']} />
+                            <FontAwesomeIcon fixedWidth icon="farStar" title="Set as display nickname" />
```
UncleClapton: the previous `['far', 'star']` value was the correct icon. This value does not work and results in an error. 
```
Could not find icon { prefix: 'fas', iconName: 'farStar' }
```
We only need to rename the export in `library.js` to deal with the naming conflict, but the name used by `<FontAwesomeIcon />` remains the same. These are values defined within the icon objects we export in `library.js`

UncleClapton (changes requested):

PR #460, opened 2025-09-02
feat: add Docker support with secrets and update CI workflow

SuperManifolds:
- Add Docker support with multi-stage build using Node 22
- Implement Docker secrets with fallback to environment variables
- Update .config/env.js to check Docker secrets first, then env vars
- Add docker-compose.yml with secrets configuration
- Create secrets directory with examples and documentation
- Update GitHub Actions workflow to include Docker build and test
- Add Docker caching for faster CI builds
- Test container startup and HTTP response in CI

## Description
Implemented a dockerfile for the fuelrats website as part of an effort to containerise all our services 

### Contributor's Checklist

* [x] This PR was created to resolve an existing issue or set of issues.
* [x] This PR satisfies any and all acceptance criteria laid out by issue(s) it resolves.
* [x] I have discussed creating this PR with the maintainers in the issue(s) beforehand.
* [x] I have thoroughly tested the changes this PR introduces in a local development environment.
* [x] I have linted the entire codebase using `yarn lint` and confirmed there are no errors or warnings.
* [x] I have followed the commit conventions laid out by [`CONTRIBUTING.md`](https://github.com/FuelRats/fuelrats.com/blob/develop/CONTRIBUTING.md#commit-conventions).

### Maintainer's Checklist

* [ ] User-facing changes this PR introduces have been captured in a changeset.
* [ ] This PR has been linted and tested locally as a part of the review process.
* [ ] All issues this PR resolves have been properly linked before merging.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -39,19 +39,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
-
       - name: Build Docker image
-        uses: docker/build-push-action@v5
-        with:
-          context: .
-          push: false
-          tags: fuelrats-web:pr-${{ github.event.pull_request.number }}
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
-          build-args: |
-            NODE_ENV=production
+        run: docker build -t fuelrats-web:test .
 
       - name: Test Docker container starts
         run: |
@@ -60,7 +49,7 @@ jobs:
             -p 3000:3000 \
             -e NODE_ENV=production \
             -e APP_URL=http://localhost:3000 \
-            fuelrats-web:pr-${{ github.event.pull_request.number }}
+            fuelrats-web:test
           
           # Wait for container to be ready
           sleep 10
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -32,9 +32,55 @@ jobs:
           stylelint_args: "--max-warnings 0"
           stylelint_extensions: "css,sass,scss"
 
+  build-docker:
+    runs-on: ubuntu-latest
+    name: Build and Test (Docker)
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build Docker image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          push: false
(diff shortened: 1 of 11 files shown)


Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+COPY . .
+
+# Build the Next.js application
+RUN yarn build
+
+# Production stage
+FROM node:22-alpine AS runner
+WORKDIR /app
+
+ENV NODE_ENV production
```
UncleClapton: I forget how these files execute. Would this ENV be set for the `yarn build` step above? We want this var set at build time, but _only_ for production mode.

.github/workflows/pull_request.yml
```diff
@@ -32,9 +32,44 @@ jobs:
           stylelint_args: "--max-warnings 0"
           stylelint_extensions: "css,sass,scss"
 
+  build-docker:
```
UncleClapton, line 35: Any reason why we would only want to test in pull requests? I think if we're containerizing on the server as well we should double check before deployment too.

docker-compose.yml
```diff
@@ -0,0 +1,70 @@
...
+  web:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    ports:
+      - "3000:3000"
+    environment:
+      # Application settings
+      - NODE_ENV=production
+      - PORT=3000
```
UncleClapton: Would there be a way to set this up for development in a way that doesn't require us to hard code `NODE_ENV` and `PORT`?

UncleClapton (changes requested):

.github/workflows/pull_request.yml
SuperManifolds: I had neglected to modify the publish workflow to include docker stuff because the existing publish workflow will probably be entirely replaced when we move to docker in prod, but I've added a test now

Dockerfile
SuperManifolds: Updated the dockerfile to make this more configurable

docker-compose.yml
SuperManifolds: Updated the dockerfile to make this more configurable

Dockerfile
UncleClapton: Just to confirm before i accept this, the current revision will set `NODE_ENV=production` when docker-compose runs `yarn build` for production right? Without this we'd be deploying an unoptimized build.

SuperManifolds: I've adjusted it so NODE_ENV can be passed from environment during the build stage, it is not currently set anywhere since we do not have a docker based deployment workflow yet, this will have to come in a later PR when we do the actual move of prod.

UncleClapton (approved):
LGTM! Thanks :)

